PR #1, opened 2016-09-26
Removed reference to Bitbucket and replaced it with Github

soerenjalas:
Just a small typo...


RemiLehe:
Thanks for noticing!
I just saw that openPMD-viewer has the same error!

PR #5, opened 2016-10-14
Allow each MPI rank to perform a separate simulation

RemiLehe:
When running parametric cases, it would nice to be able to run several cases simultaneously **with a single fbpic_script** launched as a single MPI job, whereby each MPI rank runs a separate, one-proc simulation.

Of course, one alternative is to launch each cases as a separate one-proc job. However, launching many one-proc job has disadvantages compared to launching a single, large MPI job which itself runs several independent simulations (one per MPI rank):

- Launching many one-proc jobs does not allow to e.g. run 4 independent simulations simultaneously on a node with 4 K80 GPUs. (since the different jobs cannot share a same node)
- On some clusters, there is apparently a limit for the number of small jobs that can run simultaneously (e.g. only 2 jobs on Titan). Thus, running e.g. 20 small jobs will take a long time, since only 2 jobs can run at a time., whereas running a single job with 20 MPI ranks (performing separate simulations) is much faster.

For the above reasons, I added a flag `use_all_mpi_ranks` for the `Simulation` object. When this is set to True (default), the simulation is performed on all nodes using domain-decomposition, but when it is set to False, each MPI rank performs an independent simulation, thereby allowing to do parameter scans in parallel. NB: this assumes that, in the fbpic_script.py, the input parameters of the simulation depend on the MPI rank: see the example here:

```
...

from mpi4py.MPI import COMM_WORLD as comm

# The laser
if comm.rank == 0:
    a0 = 4.          # Laser amplitude
else:
    a0 = 2.
w0 = 5.e-6       # Laser waist
ctau = 5.e-6     # Laser duration
z0 = 15.e-6      # Laser centroid

...
```
Also, the output directory should also depend on the MPI rank, like here:
```
    # Add a field diagnostic
    write_dir = 'proc%d' %comm.rank
    sim.diags = [ FieldDiagnostic( diag_period, sim.fld,
                    comm=sim.comm, write_dir=write_dir ),
                ParticleDiagnostic( diag_period,
                    {"electrons" : sim.ptcl[0]}, select={"uz" : [1., None ]},
                    comm=sim.comm, write_dir=write_dir ) ]
```

In order to indeed allow different MPI ranks to perform separate simulations, I have set their `BoundaryCommunicator` so that:

- its size is 1
- its MPI communicator is `None` (so as to be sure that the separate MPI simulations do not communicate)

This implied several changes to the code:

- Since the size of the `BoundaryCommunicator` and the `mpi4py` communicator can thus be different, I had to make sure that an fbpic simulation always looks only at the size of the `BoundaryCommunicator`
- For size = 1, I had to prevent all the calls to MPI gather (before, it was still called, even when doing a one-proc simulation).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ before_install:
 # Install packages
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION numba matplotlib scipy h5py
+  - conda install --yes gcc 
   - conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION pyfftw mpi4py
   - conda install --yes pyflakes
   - pip install openpmd_viewer
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ before_install:
 # Install packages
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION numba matplotlib scipy h5py
-  - conda install --yes gcc 
   - conda install --yes -c conda-forge python=$TRAVIS_PYTHON_VERSION pyfftw mpi4py
   - conda install --yes pyflakes
   - pip install openpmd_viewer
--- a/docs/source/example_input/lwfa_script.py
+++ b/docs/source/example_input/lwfa_script.py
@@ -43,7 +43,7 @@ rmax = 20.e-6    # Length of the box along r (meters)
 Nm = 2           # Number of modes used
 # The simulation timestep
 dt = (zmax-zmin)/Nz/c   # Timestep (seconds)
-N_step = 200     # Number of iterations to perform
+N_step = 300     # Number of iterations to perform
 
 # The particles
 p_zmin = 25.e-6  # Position of the beginning of the plasma (meters)
--- a/docs/source/example_input/lwfa_script.py
+++ b/docs/source/example_input/lwfa_script.py
@@ -43,7 +43,7 @@ rmax = 20.e-6    # Length of the box along r (meters)
 Nm = 2           # Number of modes used
 # The simulation timestep
 dt = (zmax-zmin)/Nz/c   # Timestep (seconds)
-N_step = 300     # Number of iterations to perform
+N_step = 200     # Number of iterations to perform
 
 # The particles
 p_zmin = 25.e-6  # Position of the beginning of the plasma (meters)
--- a/docs/source/example_input/parametric_script.py
+++ b/docs/source/example_input/parametric_script.py
@@ -1,0 +1,128 @@
+"""
+This script runs two simulations in parallel using MPI, 
+with each simulation having a different value of a0 as input.
+(Each simulation is performed by a different MPI process.)
+
+This makes use of the parameter `use_all_mpi_ranks=False` in 
+the `Simulation` object, which allows each MPI rank to carry out a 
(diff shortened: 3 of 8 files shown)


.travis.yml
```diff
@@ -16,6 +16,7 @@ before_install:
 # Install packages
 install:
   - conda install --yes python=$TRAVIS_PYTHON_VERSION numba matplotlib scipy h5py
+  - conda install --yes gcc
```
RemiLehe: This is unrelated to the rest of the pull request, but was necessary to get the tests to pass.
(see https://github.com/conda-forge/mpi4py-feedstock/issues/1)

soerenjalas:
Hi,
Nice PR this will be very useful! Though, I think on the long term, as the possibility  to do parameter scans are one key feature of fbpic, we should think of a more handy way to insert parameter ranges.

For example do something like:

```python
a0 = [3, 4, 4.5, 1667]
```

instead of 

```python
if comm.rank == 0:
    a0=3
elif comm.rank==1:
   a0=4
...
```

MKirchen:
We can also just define the following in the input script:

```python
a0_scan = [3, 4, 4.5, 1667] 
```
and then each rank gets the correct value like this:

```python
a0 = a0_scan[comm.rank] 
```

Apart from this, as mentioned already by Sören, it would be nice to combine the parallel execution with the parallel scan feature. However, you can decide wether or not this is too complicated to implement. Might get confusing...

RemiLehe:
Ok, I like the way in which the parameters are indicated (btw, `a0 = 1667`??). How about adding an example script for parallel scans in the documentation examples (with automatic testing, as for the other example scripts) ? If it sounds good to you, I'll add it to this PR.

Regarding mixing domain-decomposition and multiple simulations, I think it would definitely possible by using MPI subgroups. However, it might be complicated for the user to give the right parameters in the input script (for instance for 6 MPI ranks, 3 different simulations and 2 ranks per simulations, you would need to do `a0_list = [3,4,5]; a0 = a0_list[ int(comm.rank/2) ]`). Thus for the moment, I would put this on hold, but I agree it would be a nice feature in the future, if we find a nice way to input the parameters.

soerenjalas:
You gotta stay ambitious concerning your lasers... Making an example file sounds good. However, we can also do this in a future PR if you prefer that.

MKirchen:
If this doesn't consume too much of your time, you can add an example file. This would be nice to have. Otherwise, I can also merge as is. Just let me know.

RemiLehe:
Sorry for my late reply: I was very busy yesterday. 
I'll add the example over the weekend ; I think it's good to have it, and to have the automatic test that make sure this feature continues to work properly.

MKirchen:
Should I merge although tests failed?

RemiLehe:
No, actually I am still working on it. 
I have now put the flag WIP (work in progress) in the title of the PR, as is common practice on Github to show that this is not ready for merging. I'll remove it once I am ready.

Sorry for the delay.

RemiLehe:
OK, now it's ready. Can you check the new example input file + the corresponding automatic test? If they seem good to you, you can merge.

MKirchen:
Looks good!

RemiLehe:
Thanks!

PR #6, opened 2016-10-20
Incorporate the Galilean scheme

RemiLehe:
There are several additional changes that I would like to make on the code for the Galilean scheme. However, this current pull request is getting very big, and somewhat difficult to review. So I think that:
- This pull request can now be merged (The main change that I introduced recently is that I added an automatic test which checks that the Galilean scheme prevents the instability)
- I will propose additional change in the `Issues` section and implement them in separate pull requests.


MKirchen:
Looks good! I will merge this. Also, I agree with changing a few things later to make the code cleaner.

PR #7, opened 2016-10-20
Avoid importing matplotlib in production runs

RemiLehe:
Before this PR, matplotlib was imported by default by fbpic (in `fields.py`). However, it is only used for some plots of the fields, which were used when developing the code, but are never used in production.

Since the latest version of matplotlib take quite some time to be imported, this PR stops importing matplotlib by default (so as not to slow down production runs). matplotlib is now only imported when using the above-mentioned plotting functions.

As a consequence, matplotlib was also removed from the dependencies of fbpic (and from the corresponding documentation). It is now only installed for running the tests (since the tests do still use matplotlib)

PS: Either of the three of you can merge this PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,6 @@ install:
   - python setup.py install
 before_script:
   - python -m pyflakes .
-script: 
+  - pip install matplotlib openPMD-viewer
+script:
   - "python setup.py test"
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,10 +15,9 @@ before_install:
 
 # Install packages
 install:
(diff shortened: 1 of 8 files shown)


MKirchen:
Should we maybe introduce a "try -> except" for the import of matplotlib if its not a dependency of FBPIC anymore? That way, a user would be notified that he needs to install matplotlib if he wants to call this function.

RemiLehe:
Sounds good. I think we have a similar mechanism for `openPMD-viewer`, which is not a strict dependency of fbpic, but is however required for the restarts.

RemiLehe:
The PR is ready for merging :)

MKirchen:
Cool, thanks for taking care of that small but important detail.

PR #15, opened 2016-11-09
Minor changes to the galilean documentation

RemiLehe:
This includes two small changes:
- Remove the note in the documentation that said that the Galilean scheme is not implemented.
- By default, do not output rho and J in the boosted field diagnostic.

The second change is not ideal, but it is motivated by the fact that I often observed rho and J to be quite noisy and hard to interpret in the backward-transformed diagnostics (whereas they are fine in the untransformed diagnostics, i.e. the diagnostics within the boosted frame). This is clearly an artifact of the backward transformation for the staggered field rho and J. 
So the current default is a safe-guard for new users, so that they do not get those weird fields and wonder what happens.  Of course, they can always override this default, but in this case they will (hopefully) read the corresponding documentation/docstring and be aware of the potential issues.

@MKirchen : do you agree with this?

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -21,7 +21,7 @@ algorithm, its advantages and limitations, see the section :doc:`overview`.
 
 FBPIC can run either on CPU or GPU. For large
 simulations, running the code on GPU can be up to **40 times faster**
-than on CPU. 
+than on CPU.
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


MKirchen:
Is the noise maybe also related to a difference in filtering the output data? Rho and J get filtered in spectral space, but I am not sure if this filtering is always also applied to the output data?

Still, I am fine with your suggested solution, as it makes sense that the staggering is responsible for noise in the output.

RemiLehe:
Yes, part of the noise is certainly due to the fact that we do not filter rho and J (at least not in real space). 
But if this was the only reason, we would also see ugly fields in the non-transformed diagnostics, which is not the case. So, this is why I think the staggering may play a role.

Issue #14, opened 2016-11-08
Build numba functions for the field push


RemiLehe:
Right now, the field push functions on the CPU are written in numpy, inside of `fields.py`. I think it could be better to have them written in numba, in a separate file. Also, it is a minor point, I would see two advantages to this:
- The functions would be much closer to the GPU functions. Since the GPU functions are not automatically tested, but the CPU functions are, this would make slightly less likely that there is a bug in the GPU functions. (Although it is by no means a guarantee.)
- The file `fields.py` would be shorter and more readable.

MKirchen:
I agree, this makes sense.

PR #20, opened 2016-11-11
Close #12 and #14: Build numba function for field push and keep the ones for standard PSATD

### LINKED ISSUE #12: Make field push more readable: keep the standard PSATD routines
In pull request #6, the field push routine for the standard PSATD is replaced by the field push routine for the Galilean/comoving scheme. 

However, I think this makes it less readable for people who are just interested in the standard PSTAD scheme. Although the Galilean/comoving coefficients reduce to the standard PSATD when `ps.V` is 0, it is not obvious at a first glance.  I think keeping a separate, simpler function for the standard PSATD would be easier.

@MKirchen : Any opinion on this?

RemiLehe:
This pull request creates numba functions for the field push and the current correction, when done on the CPU.

It also reintroduces the standard PSATD functions that were replaced by the galilean/comoving functions. This is for the sake of simplicity, for users that are only interested in the standard PSATD.

Note: I did not include the change that @MKirchen requested regarding setting V to None for the standard PSATD scheme. This is because I did not want this PR to become too big. I will include this in a future PR.

### DIFF
--- a/fbpic/fields/cuda_methods.py
+++ b/fbpic/fields/cuda_methods.py
@@ -137,29 +137,41 @@ def cuda_divide_vector_by_volume( mode0r, mode1r, mode0t, mode1t,
 # Methods of the SpectralGrid object
 # -----------------------------------
 
+@cuda.jit('void(complex128[:,:], complex128[:,:], \
+           complex128[:,:], complex128[:,:], complex128[:,:], \
+           float64[:,:], float64[:,:], float64[:,:], \
+           float64, int32, int32)')
+def cuda_correct_currents_standard( rho_prev, rho_next, Jp, Jm, Jz,
+                            kz, kr, inv_k2, inv_dt, Nz, Nr ):
+    """
+    Correct the currents in spectral space, using the standard pstad
+    """
+    # Cuda 2D grid
+    iz, ir = cuda.grid(2)
+
+    # Perform the current correction
+    if (iz < Nz) and (ir < Nr) :
+
+        # Calculate the intermediate variable F
+        F = - inv_k2[iz, ir] * (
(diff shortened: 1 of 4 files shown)


fbpic/fields/cuda_methods.py
```diff
@@ -137,29 +137,41 @@ def cuda_divide_vector_by_volume( mode0r, mode1r, mode0t, mode1t,
 # Methods of the SpectralGrid object
 # -----------------------------------
```
RemiLehe: For conciseness, I removed some of the docstring.
Unfortunately, github shows this change as mixed up with the body of the function :/

fbpic/fields/numba_methods.py
```diff
@@ -0,0 +1,209 @@
+# Copyright 2016, FBPIC contributors
```
RemiLehe: Here are the new numba functions for the CPU.

Except for the loop over iz and ir, they are indentical to the corresponding cuda functions. 
Maybe we could build a numba "inline function" to avoid code duplication.
@MKirchen, @Escapado  : I know less than you do on this. Do you have any suggestions?

fbpic/fields/fields.py
```diff
@@ -1151,11 +1138,6 @@ def __init__( self, kz, kr, m, dt, Nz, Nr, V=0.,
         else:
              self.j_corr_coef = inv_dt*np.ones_like(kz)
```
RemiLehe: Thanks to the numba functions, we do not need these temporary arrays anymore!

fbpic/fields/numba_methods.py
MKirchen: No, I don't think this would make the code able to be run on both, CPU and GPU.

PR #21, opened 2016-11-11
Set V to `None` in order to require the standard PSATD algorithm

RemiLehe:
** Please merge PR #20 first, before merging this one **

As suggested by @MKirchen, this sets V to `None` in order to require the standard PSATD algorithm.


MKirchen:
@RemiLehe Nice work!

PR #24, opened 2016-11-16
Close #23: Periodic boundaries seem broken in dev branch

### LINKED ISSUE #23: Periodic boundaries seem broken in dev branch
While trying to merge the new higher order shape implementation
into the current dev branch I noticed that errors appear close to the
boundaries in the periodic plasma wave test file:

screen shot 2016-11-15 at 18 30 58

Maybe this is related to the "save_periodic" commit ?

RemiLehe:
I just corrected the bug that @Escapado reported. 
As can be seen in the modified code, the error was due to a rather significant error in the particle exchange. 
With the new code, here is map of the fields:
figure_1
Could you double-check that you get the same results with this branch, on your side?

If yes, you can then merge this branch.

### DIFF
--- a/fbpic/boundaries/particle_buffer_handling.py
+++ b/fbpic/boundaries/particle_buffer_handling.py
@@ -23,7 +23,7 @@ def remove_outside_particles(species, fld, nguard, left_proc, right_proc):
     When the boundaries are open, only the particles that are in the
     outermost half of the guard cells are removed. The particles that
     are in the innermost half are kept.
-    
+
     Parameters
     ----------
     species: a Particles object
@@ -35,7 +35,7 @@ def remove_outside_particles(species, fld, nguard, left_proc, right_proc):
 
     nguard: int
         Number of guard cells
-        
+
     left_proc, right_proc: int or None
         Indicate whether there is a left or right processor or if the
         boundary is open (None).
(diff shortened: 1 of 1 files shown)


fbpic/boundaries/particle_buffer_handling.py
```diff
@@ -196,18 +196,28 @@ def remove_particles_gpu(species, fld, nguard, left_proc, right_proc):
...
-    i_min = prefix_sum.getitem( (nguard+fld.prefix_sum_shift)*Nr )
-    i_max = prefix_sum.getitem( (Nz-nguard+fld.prefix_sum_shift)*Nr - 1 )
-    # For the open boundaries, only the particles in the outermost
-    # half of the guard cells are removed
+    # Find the z index of the first cell for which particles are kept
     if left_proc is None:
-        # Find the index in z below which particles are removed
+        # Open boundary: particles in outermost half of guard cells are removed
         iz_min = max( int(nguard/2) + fld.prefix_sum_shift, 0 )
-        i_min = prefix_sum.getitem( iz_min * Nr )
```
RemiLehe: The error in the previous version of the code is that I forgot to add -1 in the above line, i.e. I have
```
i_min = prefix_sum.getitem( iz_min * Nr )
```
instead of 
```
i_min = prefix_sum.getitem( iz_min * Nr - 1 )
```
This meant that, in addition to the particles of the left guard cells, we also passed the particles of the **first cell of the valid physical domain** to the left proc -- which is invalid. Note that the calculation of `i_max` was correct. I don't know why I did it for `i_max` and not `i_min`.

Anyway, once I corrected this bug, I also realized this whole part would be slightly more readable if reorganize it. Sorry if that makes the PR less readable. But the real only change is the -1, the rest is refactoring.

MKirchen:
Thanks for fixing this. Was probably hard to track down - nice work!

PR #26, opened 2016-11-17
[WIP] Implement cubic particle shapes for deposition on GPU 

Escapado:
This PR is meant to implement cubic particle shapes for the deposition of rho and j on the GPU.
Additionally the deposition routine for linear shaped particles has been rewritten to use atomic operations. 

Using nvprof with the periodic plasma wave testfile on an Nvidia K20x the new kernels overall perform a little bit better, by a margin of about 5 to 10 percent and also only one kernel is needed for the deposition step.

For now cubic order particle shapes take between 4 and 5 times longer than linear order. This is to be expected since the number of deposition sites also increases four-fold.

To access the different kernels one has to specify the `particle_shape` attribute in the simulation object or when intializing particles. The valid arguments are `linear`, `cubic` and `linear_non_atomic`. The former two being the new kernels and the latter being the old kernel which was kept for testing purposes.

Please keep in mind that this is a **work in progress**.


RemiLehe:
Awesome! This is really an impressive work, and thanks a lot for sharing it!!
I'll try to review the code within the next few days if that's okay with you.

MKirchen:
Please note that the code might change again quite a lot and we still need to add the new gathering methods. However, we decided to open a PR already so that we can continuously discuss these changes.

RemiLehe:
OK, great. Thanks for letting me know.
Although it is an exception to our general rule, would you be okay with pushing the branch to the main repository (i.e. fbpic/fbpic) instead of to a fork (i.e. Escapado/fbpic)? That way, we can all modify the branch without having to open new pull requests towards the fork?

Escapado:
Yes, for me this is totally fine.

RemiLehe:
Great. I just pushed your branch to `fbpic/fbpic`. 
Could you close the present pull request, and create a new pull request from the branch `fbpic/fbpic:higher-order-shapes` to `fbpic/fbpic:dev` with the same title and introductory text? Then we can work directly on the branch `fbpic/fbpic:higher-order-shapes`.

PR #27, opened 2016-11-17
Implement cubic particle shapes for deposition on GPU

Escapado:
This PR is meant to implement cubic particle shapes for the deposition of rho and j on the GPU.
Additionally the deposition routine for linear shaped particles has been rewritten to use atomic operations. 

Using nvprof with the periodic plasma wave testfile on an Nvidia K20x the new kernels overall perform a little bit better, by a margin of about 5 to 10 percent and also only one kernel is needed for the deposition step.

For now cubic order particle shapes take between 4 and 5 times longer than linear order. This is to be expected since the number of deposition sites also increases four-fold.

To access the different kernels one has to specify the `particle_shape` attribute in the simulation object or when intializing particles. The valid arguments are `linear`, `cubic` and `linear_non_atomic`. The former two being the new kernels and the latter being the old kernel which was kept for testing purposes.

Please keep in mind that this is a **work in progress**.


RemiLehe:
Again thanks a lot for this work!

For readability, I think that it would be worth it to split the `cuda_methods,py` into several files (this would also (hopefully) allow it to be displayed in Github). For instance, the structure of the `particles` folder could be as follows:
```
__init__.py
cuda_methods.py
numa_methods.py
particles.py
utility_methods.py
cuda_deposition/
      linear_non_atomic.py
      linear.py
      cubic.py
````
where `cuda_methods.py` would contain all the functions for particle push, gathering, sorting, etc., and where `cuda_deposition/` would contain the functions related to the charge/current deposition.

Would this make sense? I have not read the code in details, so maybe the above suggestion is not really advantageous... Feel free to reject it if :)

**Comment by RemiLehe** [comment]
Yes, doing some benchmarking and exposing the results would be a great idea!

Also, we should probably also include the use of the 3rd order shape (at least on CPU) in the existing automated tests. I can do this, if it is okay with you.

**Comment by MKirchen** [comment]
As we discussed today: @Escapado  Is this ready for review by @RemiLehe ?

Could you write a quick summary of what we did recently and what to conclude on this?
Or do we need to take care of some more details before merging?

**Comment by Escapado** [comment]
The PR is ready for review.

Here is a quick summary of the strategy behind the changes. As kernel performance is a little unpredictable under the change of code bits we tested a big set of different implementation details. 

For the deposition kernels it turned out that unrolling loops and using registers for the different local values which are atomically added resulted in a significant performance gain, compared to using local arrays and loops. Also using atomic operations isn't slow anymore. In fact the new approach is even slightly faster in linear order than the old implementation and no additional kernels are needed to add things together as before. Admittedly this makes the kernels less readable which is why the deposition kernels, especially for cubic order, were put into separate files (as suggested by @RemiLehe ).

For gathering this approach was also tested. However unrolling loops resulted in a significant slowdown of the kernels. The same goes for using registers of the shape factors instead of using cuda.local.arrays. Also inlining a lot of operations has been extensively tested, (for example Fi_m were discarded and operations were put into one line or shapes were computed inline etc).  Even something as simple as replacing
```
invr = 1./rj
cos = xj*invr  # Cosine
sin = yj*invr  # Sine
```
by 
```
cos = xj/rj  # Cosine
sin = yj/rj  # Sine
```
can alter performance for the worse sometimes even though one would expect a speed up since we are saving one register. However I think all this boils down to the compiler optimising code on it's own in a non-transparent way.

It turns out that for cubic order the fastest implementation we could find is the one that is implemented in the branch right now. For linear particle shapes the original implementation from @MKirchen is still the fastest which is why it is left unchanged in the codebase.

RemiLehe:
Great thanks for all this information. I'll try to review this PR soon.

Thanks again for your great work!

RemiLehe:
Hi Kevin,
I have started reviewing the pull request. For the moment, I have reviewed all files except `linear_non_atomic.py`, `linear.py` and `cubic.py`. 

I'll do a pull request in the next hour to the branch `higher-order-shapes` with proposed changes to the files that I have reviewed. 
Then I'll do a second pull request later with proposed changes to the files `linear_non_atomic.py`, `linear.py` and `cubic.py`.

Once you have accepted/edited these two pull requests, we can merge the branch `higher-order-shapes` into the branch `master`.

Thanks again for your nice work!

RemiLehe:
I finally got the time to test this PR on GPU (and also, `accelerate` is now compatible with `numba0.27` and later versions, which allows us to use this PR in production).

Everything works fine and this is ready for merging. @Escapado : many thanks again for this awesome PR, and the new functionality that it brings !!

PR #28, opened 2016-12-09
Remove `inv_gamma` from the `slice_array` in the boosted diagnostics

RemiLehe:
I am currently working on implementing the boosted diagnostics for multi-CPU and multi-GPU simulations, as it is currently needed for one of our users. (Note that the regular diagnostics are implemented for multi-CPU/GPU simulations (via MPI gather to the first proc), but when we implemented the boosted diagnostics, we skipped it for the sake of rapidity of the implementation).

The current pull request does not implement this directly, but is a minor change in preparation for it.

In the current pull request, I slightly modify the structure of the data. Throughout the different operations that we do on the data (interpolation at a given time, Lorentz-transform, selection of particles, etc.) we use arrays of size (8, num_particles) where the 8 corresponds to the number of particle quantities: x, y, z, ux, uy, uz, inv_gamma, w.

However, inv_gamma is only needed to do the interpolation at a given time and the Lorentz-transform, but then we can drop it, so that the array that we send/receive via MPI gather to the first proc can be only of size (7, num_particles). This is what the current pull request does: I keep a dictionary structure before the Lorentz transformation, and then after it, I switch to a 2darray structure of shape (7, num_particles).

### DIFF
--- a/fbpic/openpmd_diag/boosted_particle_diag.py
+++ b/fbpic/openpmd_diag/boosted_particle_diag.py
@@ -495,8 +495,7 @@ class ParticleCatcher:
         # Create a dictionary that contains the correspondance
         # between the particles quantity and array index
         self.particle_to_index = {'x':0, 'y':1, 'z':2,
-                                  'ux':3,'uy':4, 'uz':5,
-                                  'inv_gamma':6, 'w':7}
+                                  'ux':3,'uy':4, 'uz':5, 'w':6}
 
     def extract_slice( self, species, current_z_boost, previous_z_boost,
                        t, select=None ):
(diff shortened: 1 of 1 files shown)


RemiLehe:
My text editor automatically removed white spaces and empty lines, which makes the PR difficult to review. However, I separated the white-spaces changes and the true changes into 2 different commits. 

So for reviewing, just look at the second commit (using the commits button above)

fbpic/openpmd_diag/boosted_particle_diag.py
```diff
@@ -748,19 +744,29 @@ def interpolate_particles_to_lab_frame( self, slice_array, current_z_boost, t ):
         z_lab = self.gamma_boost*( z + self.beta_boost*c*t_cross )
 
         # Back-transformation of momentum
-        gamma = np.sqrt(1. + (ux**2 + uy**2 + uz**2))
+        gamma = 1./inv_gamma
```
RemiLehe: The above change is not crucial here (both operations are equivalent), but I guess the new one is slightly faster.

PR #29, opened 2016-12-09
Parallel boosted frame output

RemiLehe:
This generalizes the boosted-frame output to the case of a multi-CPU or multi-GPU.

The implementation is similar to that of [Warp](https://bitbucket.org/berkeleylab/warp).

More specifically, the data that needs to be written is cached on each processor until the simulation reaches an output iteration (as defined by the parameter `diag_period`). Then this cached data is gathered on rank 0, and rank 0 writes the data to disk.

Note: I also modified the boosted-frame example input script (to make it faster to run), and switched the boosted-frame automated test from single proc to two-proc. (So that the parallel boosted frame output are automatically tested for each push.)

### DIFF
--- a/fbpic/openpmd_diag/boosted_particle_diag.py
+++ b/fbpic/openpmd_diag/boosted_particle_diag.py
@@ -210,7 +210,7 @@ class BoostedParticleDiagnostic(ParticleDiagnostic):
 
                 snapshot.buffered_slices[species_name] = []
 
-    def gather_particle_arrays( array, root=0 ):
+    def gather_particle_arrays( self, array, root=0 ):
         """
         Gather the compacted arrays of particle slices, on the
 
@@ -230,7 +230,7 @@ class BoostedParticleDiagnostic(ParticleDiagnostic):
         """
         # Send the local number of particles to all procs
         n_particles_local = array.shape[1]
-        n_particles_list = self.comm.allgather( n_particles_local )
+        n_particles_list = self.comm.mpi_comm.allgather( n_particles_local )
 
         # Prepare the send and receive buffers
         if self.rank == root:
@@ -246,7 +246,7 @@ class BoostedParticleDiagnostic(ParticleDiagnostic):
         recvbuf = [ gathered_array, n_size_procs, i_start_procs, REAL8 ]
 
         # Send/receive the arrays
-        self.mpi_comm.Gatherv( sendbuf, recvbuf, root=root )
+        self.comm.mpi_comm.Gatherv( sendbuf, recvbuf, root=root )
 
         # Return the receive buffer
         return( recvbuf )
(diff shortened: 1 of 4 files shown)


fbpic/openpmd_diag/boosted_particle_diag.py
```diff
@@ -190,18 +191,66 @@ def flush_to_disk(self):
             # over time into a single array
             for species_name in self.species_dict:
 
-                if snapshot.buffered_slices[species_name] != []:
```
RemiLehe: Note: I had to get rid of the above `if` condition. Otherwise, if one proc does not have any buffered slices for this snapshot (but other procs do), then this proc will not participate in the MPI gather below, and the other procs will just for the mpi gather indefinitely.

fbpic/openpmd_diag/boosted_particle_diag.py
```diff
@@ -459,13 +508,15 @@ def compact_slices( self, species ):
         -------
         paticle_array: an array of reals of shape (7, numPart)
         regardless of the dimension
-
-        Returns None if the slices are empty
+        Returns an array of size (7,0) if the slices are empty
```
RemiLehe: This modification is also made so that a proc which has no data at all can still participate in the MPI gather (by sending an empty array).

MKirchen:
Nice implementation. One small remark:

I know that you needed to use the slow gather routine as we can not know which process has some data and which has not. However, after thinking about it, maybe it would still be nice to directly implement the fast gather routine everywhere. One possibility would be to use Gatherv, while exchanging the shape information about the arrays with a slow routine before:

`Gatherv(self, sendbuf, recvbuf, int root=0)
Gather Vector, gather data to one process from all other processes in a group providing different amount of data and displacements at the receiving sides`

Another possibility would be to use the normal (fast) Gather while using an MPI subgroup consisting of all MPI processes but the ones that have no slices.

Do you think thats feasible? Maybe its not needed as we do only gather every few hundred time steps. Its your decision. We can also leave it like this for now and if we see that this affects performance when running benchmarks later, we can still change it.

I am ready to merge if you like.

RemiLehe:
I think that using `Gather` would be a good idea. However, instead of using an MPI subgroup (which might induce considerable overhead when being created) I would just use `COMM_WORLD` and modify the code so that the procs which used to send `None` would now send an empty array.

That being said, I will probably not have the time to implement the above anytime soon. If you are interested in implementing this, please go ahead on the same branch. Otherwise, you can just merge. 

Thanks for review, by the way :)

MKirchen:
I will merge for now and we can do some benchmarks in the future. If it turns out that it may affect performance, I will implement the fast routines.

Thanks again for implementing this!

RemiLehe:
Great! Thanks for merging.

PR #31, opened 2017-01-04
Proposed changes to 3rd order implementation in all files except `linear.py` and `cubic.py`

RemiLehe:
Hi Kevin,

Here are a few suggested changes to the pull request on 3rd order shape factor (except for the files `linear.py` and `cubic.py`, for which I will propose changes in a separate pull request).

Feel free to reject/discuss some of these changes, they are only suggestions.

Here are the main changes:
- I added you name to the files you changed (in the header)
- I rephrased some of the docstrings and some of the inline comments (this is me being very picky ; sorry!)
- In the CPU function `weights`, I used 2darrays directly instead of lists of 1darrays which are later converted to 2darrays. I am not sure it is much faster, but it makes the implementation slightly more compact.
- I reverted the changes that I made to the file `test_laser.py` (I previously added tests that initialize the `Simulation` object with different shape order and then test the laser antenna, but when reviewing your PR in more details, I realized that the laser antenna always uses shape factors of order 1, so there is no point in changing the shape factors for the `Simulation` object)

### DIFF
--- a/fbpic/particles/cuda_methods.py
+++ b/fbpic/particles/cuda_methods.py
@@ -1,5 +1,5 @@
 # Copyright 2016, FBPIC contributors
-# Authors: Remi Lehe, Manuel Kirchen
+# Authors: Remi Lehe, Manuel Kirchen, Kevin Peters
 # License: 3-Clause-BSD-LBNL
 """
 This file is part of the Fourier-Bessel Particle-In-Cell code (FB-PIC)
--- a/fbpic/particles/cuda_methods.py
+++ b/fbpic/particles/cuda_methods.py
@@ -482,7 +482,7 @@ def gather_field_gpu_cubic(x, y, z,
     """
     Gathering of the fields (E and B) using numba on the GPU.
     Iterates over the particles, calculates the weighted amount
-    of fields acting on each particle based on its shape (linear).
+    of fields acting on each particle based on its shape (cubic).
     Fields are gathered in cylindrical coordinates and then
     transformed to cartesian coordinates.
     Supports only mode 0 and 1.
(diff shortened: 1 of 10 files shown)


fbpic/particles/cuda_methods.py
```diff
@@ -609,9 +609,11 @@ def gather_field_gpu_cubic(x, y, z,
                 Fr_m += Sz[index_z]*Sr[index_r]*Er_m0[iz[index_z], ir[index_r]]
                 Ft_m += Sz[index_z]*Sr[index_r]*Et_m0[iz[index_z], ir[index_r]]
                 if Sz[index_z]*Sr[index_r] < 0:
-                    Fz_m += (-1.)*Sz[index_z]*Sr[index_r]*Ez_m0[iz[index_z], ir[index_r]]
+                    Fz_m += (-1.)*Sz[index_z]*Sr[index_r]* \
+                        Ez_m0[iz[index_z], ir[index_r]]
                 else:
-                    Fz_m += Sz[index_z]*Sr[index_r]*Ez_m0[iz[index_z], ir[index_r]]
+                    Fz_m += Sz[index_z]*Sr[index_r]* \
+                        Ez_m0[iz[index_z], ir[index_r]]
```
RemiLehe: The above changes are just there in order to avoid having more than 80 characters per line.

fbpic/particles/numba_methods.py
```diff
@@ -206,12 +201,18 @@ def deposit_field_numba(Fptcl, Fgrid,
...
         # Loop over adjacent cells (given by shape order)
+        # Use helper variables `ir_corr` and `Sr_corr`, in order to avoid
+        # modifying ir and Sr in place. (This is not strictly necessary,
+        # but is just here as a safeguard.)
         for cell_index_r in range(ir.shape[0]):
             for cell_index_z in range(iz.shape[0]):
                 # Correct the guard cell index and sign
-                if ir[cell_index_r][ip] < 0:
-                    ir[cell_index_r][ip] = abs(ir[cell_index_r][ip]) - 1
-                    Sr[cell_index_r][ip] = sign_guards*Sr[cell_index_r][ip]
```
RemiLehe: I agree that modifying `ir` and `Sr` in-place works, but I would prefer not doing it, since it could be a somewhat unexpected side-effect of this function for users that don't read the function in details. 
Let me know if using `ir_corr` and `Sr_corr` here is fine with your.

fbpic/particles/numba_methods.py
```diff
@@ -133,29 +133,24 @@ def gather_field_numba(exptheta, m, Fgrid, Fptcl,
...
                 else:
-                    ir_corr = ir[cell_index_r][ip]
-                    Sr_corr = Sr[cell_index_r][ip]
+                    ir_corr = ir[cell_index_r, ip]
+                    Sr_corr = Sr[cell_index_r, ip]
                 # Gather the field value at the respective grid point
-                F += Sz[cell_index_z][ip] * Sr_corr * \
-                    Fgrid[iz[cell_index_z][ip], ir_corr]
+                F += Sz[cell_index_z, ip] * Sr_corr * \
+                    Fgrid[ iz[cell_index_z, ip], ir_corr]
```
RemiLehe: The above changes are just esthetical, and are here for consistency with the rest of the code: we usually use `,` instead of `][` when accessing 2darrays. Sorry again for being annoyingly picky!!

Escapado:
I agree with all of the proposed changes and I think it's important we stay consistent with the codebase so thanks for taking care of that!

RemiLehe:
Great! 
You can click on `Merge pull request` then, so that these changes get incorporated into the branch `higher-order-shapes` which is on the fbpic main repository.

PR #33, opened 2017-01-17
Add method to load particle distribution from openPMD hdf5 file

soerenjalas:
I thought now that we have this nice unified file format we might also want to leverage it as an input format for FBPIC. Therefore, I wrote and additional `add_elec_bunch` method which loads a particle distribution from an openPMD time-series and initialises it with corresponding space charge fields.
Just like the checkpoint method it uses the openPMD-viewer. 
Let me know if you think this is useful or not or if you want any changes made to the code.

### DIFF
--- a/fbpic/lpa_utils/bunch.py
+++ b/fbpic/lpa_utils/bunch.py
@@ -310,6 +310,109 @@ def add_elec_bunch_file( sim, filename, Q_tot, z_off=0., boost=None,
                              filter_currents, check_gaminv=False,
                              direction = direction)
 
+def add_elec_bunch_openPMD( sim, ts_path, z_off=0., species=None, select=None,
+                            iteration=None, boost=None, filter_currents=True):
+    """
+    Introduce a relativistic electron bunch in the simulation,
+    along with its space charge field, loading particles from an openPMD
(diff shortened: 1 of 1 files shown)


fbpic/lpa_utils/bunch.py
```diff
@@ -310,6 +310,107 @@ def add_elec_bunch_file( sim, filename, Q_tot, z_off=0., boost=None,
...
+    select: dict, optional
+        Either None or a dictionary of rules
+        to select the particles, of the form
+        'x' : [-4., 10.]   (Particles having x between -4 and 10 microns)
+        'ux' : [-0.1, 0.1] (Particles having ux between -0.1 and 0.1 mc)
+        'uz' : [5., None]  (Particles with uz above 5 mc)
+
+    iteration: integer (optional)
+       The iteration number of the checkpoint from which to restart
+       If None, the last iteration of the timeseries will be used.
```
RemiLehe: I would replace the first line by:
```
The iteration number of the openPMD file from which to extract the particles.
```

soerenjalas: Did not catch that when stealing your work. ;)

PR #34, opened 2017-02-07
OpenPMD diagnostics: specify min and max iterations

RemiLehe:
This is a very minor pull request: for a `ParticleDiagnostic` or `FieldDiagnostic` instance, it allows the user to specify the min and max iterations between which this diagnostic should be active. 

A typical use case would be to produce diagnostics with a higher output frequency at a specific moment of the simulation (e.g. when injection happens). In order to do so, the user would create two instances of e.g. `ParticleDiagnostic`:
- One with low output frequency (i.e. high `period`)
- One with high output frequency and specific `iteration_min` and `iteration_max`


MKirchen:
Okay :)

PR #35, opened 2017-02-07
Particle tracking

RemiLehe:
## Overview

This pull request adds particle tracking to `fbpic`: when activated, a unique ID is attributed to each macroparticles (including new macroparticles created by the moving window) and these ID are written in the openPMD file, so that particles can be tracked in postprocessing, using the openPMD-viewer (see https://github.com/openPMD/openPMD-viewer/pull/139)

This pull request is compatible with:
- the GPU implementation (the array of particle ID is updated whenever the particles are sorted)
- MPI communications (the arrays of particle ID are communicated via MPI)
- Checkpoint / restarts (when restarting from a checkpoint that contains particle ID, the simulation automatically loads these particle ID and tracks the corresponding species)

## API

In order to activate tracking for a given species, use:
```
species.track( comm )
```
where `species` is the `fbpic.Particles` object that represents the species to be tracked and `comm` is the `BoundaryCommunicator` (see below for why it is necessary to pass it). The API is not ideal, but works for now.

When the species is tracked, its ID is automatically added to the openPMD output (no need to pass "id" in the API of the openPMD diagnostics).

## Implementation

When a species is tracked, it has an attribute `tracker`, which is an instance of `ParticleTracker` and is responsible for:
- storing the `id` of the existing particles in an array
- generating new unique `id` whenever needed.

In order to generate ids that are unique **across all MPI processes**, the `ParticleTracker` on a given process generates integers of the form:
```
mpi_rank + k * mpi_size
```
where `mpi_rank` is the rank of this MPI process, and where `k` is an integer (which varies from macroparticle to macroparticle)

Note that because the `id` is an array of integer (whereas other particle attributes are arrays of floats), I had to create new MPI buffers and sorting buffers to work with the `id`.

### DIFF
--- a/docs/source/example_input/lwfa_script.py
+++ b/docs/source/example_input/lwfa_script.py
@@ -69,13 +69,14 @@ diag_period = 10         # Period of the diagnostics in number of timesteps
 save_checkpoints = False # Whether to write checkpoint files
 checkpoint_period = 50   # Period for writing the checkpoints
 use_restart = False      # Whether to restart from a previous checkpoint
+track_electrons = False  # Whether to track and write particle ids
 
 # The density profile
 ramp_start = 30.e-6
 ramp_length = 40.e-6
 
 def dens_func( z, r ) :
-    """Returns relative density at position z and r"""    
+    """Returns relative density at position z and r"""
     # Allocate relative density
     n = np.ones_like(z)
     # Make linear ramp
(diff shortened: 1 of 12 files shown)


MKirchen:
Nice work!

However, I don't know if maybe the separation between "float" and "integer" is difficult to understand for someone else. It maybe would have been nicer if we would have exchanged something like (N_left, dtypes), where dtypes is a list of datatypes of the buffers and thereby keeping it general instead of hardcoding the two datatypes float and uint directly.

However, maybe there are reasons that I don't understand, which led you to implement it this way.

Anyway, I would be ready to merge, unless you want to comment on this. I still think that the implementation is perfectly fine as is.

RemiLehe:
Hi Manuel,
Thanks for reviewing the code.

Regarding the MPI exchange part: the reason I wanted to separate the float and integer part is because we typically pack the different particle quantities into one array (that necessarily is of one single type) before doing the MPI exchanges. We are doing this packing for performance reasons.

If we had a list of quantities with their `dtype`, we would need to do separate MPI calls for each quantity (i.e. roughly 8 or 9 MPI calls), instead of packing all the floats together and all the integer together and then doing only 2 MPI calls. 
(Note: I'm saying "packing all the integer quantity together" because, even though we have only one integer quantity right now - the particle `id`, we will have another integer quantity - the ionization level - very soon.)

Does the above make sense? Did I actually understand your question correctly?

MKirchen:
Ah okay, yes that makes sense. I somewhat thought that also the "packing" could have been done more generally in a "cleaner" ways concerning the amount of code needed. But the more I think about it, the more I get to the point that your implementation makes totally sense.

Merged!

PR #37, opened 2017-02-08
Close #36 Checkpoints not working with use_all_mpi_ranks = False

### LINKED ISSUE #36: Checkpoints not working with use_all_mpi_ranks = True
Hi,
Currently it is not possible to use `use_all_mpi_ranks=True` and use checkpoints.
The problem is that `use_all_mpi_ranks=True` causes the Simulation object to set `self.mpi_comm=None`. However, the checkpoint routines try to call `sim.comm.mpi_comm.Barrier()`, which is not defined. This is used there to wait for rank 0 while it's creating directories and checking files.

I'm not too familiar with this MPI stuff but I think we could either create a communicator even when using use_all_mpi_ranks = True, or have every rank check its own files and create its own directory.

RemiLehe:
As reported by @soerenjalas, the checkpoints were not working when `use_all_mpi_ranks=False`, because the `BoundaryCommunicator` object does not contain an MPI communicator (in order to prevent the independent simulations from talking to each other).
However, when writing the checkpoints, the independent simulations do need to synchronize, in order to prepare the output folders. I fixed the issue by using the global MPI communicator **only** in the checkpoint code (see the file `checkpoint_restart.py`).

I also expanded the automatic tests so that they test the checkpoint/restart in the case of `use_all_mpi_ranks=False`.

The list of changes for this pull request is somewhat messy, because (as usual) Atom linted the code to remove trailing spaces and empty lines. Sorry about this. The only important changes are in `checkpoint_restart.py`. The rest is linting + expansion of the automatic tests.

### DIFF
--- a/docs/source/example_input/boosted_frame_script.py
+++ b/docs/source/example_input/boosted_frame_script.py
@@ -44,7 +44,7 @@ n_guard = 40     # Number of guard cells
 exchange_period = 10
 # The simulation timestep
 dt = (zmax-zmin)/Nz/c   # Timestep (seconds)
-N_step = 201     # Number of iterations to perform
+N_step = 101     # Number of iterations to perform
                  # (increase this number for a real simulation)
 
 # Boosted frame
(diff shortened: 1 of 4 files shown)


soerenjalas:
Thanks for the quick fix!

PR #40, opened 2017-02-27
Close #39: Checkpoints with external beams

### LINKED ISSUE #39: Checkpoints with external beams
This is minor and maybe doesn't even require any changes, but I want to bring it up anyways as I just ran into it.

The example input file only initialises the laser on the first start like:
```python
# Load initial fields
    if use_restart is False:
        # Add a laser to the fields of the simulation
        add_laser( sim, a0, w0, ctau, z0 )
    else:
        # Load the fields and particles from the latest checkpoint file
        restart_from_checkpoint( sim )
```

I needed to use an external electron beam instead of a laser so I just naively replaced `add_laser()` by `add_elec_bunch()`. This is fine for the first start, but leads to problems when restarting from a checkpoint. This is because `restart_from_checkpoint()` only loads particle data that is in `sim.ptcl`, which only contains the plasma particles when jumping into the `else` statement. A workaround is to just reinitialise the particle beam and than have `restart_from_checkpoint()` overwrite the data.
My suggestion would be to maybe have `restart_from_checkpoint()` add a fresh `Particle` object for every species in the checkpoint openPMD file. 
Another way to avoid this confusion could also be to just change the example file to something like

```python
    # Load initial fields
    # Add a laser to the fields of the simulation
    add_laser( sim, a0, w0, ctau, z0 )
    if use_restart is True:
        # Load the fields and particles from the latest checkpoint file
        restart_from_checkpoint( sim )
```
This way the setup is alway the same and `restart_from_checkpoint()` just updates to the current iteration.

soerenjalas:
As discussed here is a pull request to close issue #39.

### DIFF
--- a/docs/source/example_input/lwfa_script.py
+++ b/docs/source/example_input/lwfa_script.py
@@ -100,9 +100,10 @@ if __name__ == '__main__':
         use_cuda=use_cuda )
 
     # Load initial fields
+    # Add a laser to the fields of the simulation
+    add_laser( sim, a0, w0, ctau, z0 )
+
     if use_restart is False:
-        # Add a laser to the fields of the simulation
-        add_laser( sim, a0, w0, ctau, z0 )
         # Track electrons if required (species 0 correspond to the electrons)
         if track_electrons:
             sim.ptcl[0].track( sim.comm )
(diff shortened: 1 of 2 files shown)


RemiLehe:
Great! Thanks for fixing this!

PR #42, opened 2017-03-21
Fix gaussian beam initialization for zero energy spread and update test file

Escapado:
The function `np.random.normal(gamma, sig_gamma, N)` will throw a ValueError if sig_gamma<=0.
However a user might desire to initialise a bunch with a constant energy so this PR is meant to implement that by simply filling a numpy array of desired size with the desired value for each entry by using [numpy.full()](https://docs.scipy.org/doc/numpy/reference/generated/numpy.full.html). If the user selected a negative energy spread it will be set to zero and a warning is given.

Furthermore I updated the corresponding test file as it threw errors at me because Python 3.5 handles integer division differently than Python 2.x.


RemiLehe:
Great! Thanks for this PR!

PR #46, opened 2017-04-05
Simplify laser profile function and add temporal chirp

RemiLehe:
This simplifies the expression for a gaussian laser pulse:
- The expression uses complex numbers, which makes it simpler and more compact.
- In the boosted-frame case, we convert the points at which to calculate the laser to the lab-frame, instead of converting the wavelength, radius of curvature etc. to the boosted-frame.

In addition, I added the possibility to add temporal chirp (and a CEP phase) in the laser.

### DIFF
--- a/fbpic/lpa_utils/boosted_frame.py
+++ b/fbpic/lpa_utils/boosted_frame.py
@@ -219,28 +219,6 @@ class BoostConverter( object ):
 
         return( boosted_frame_vars )
 
-
-    def curvature( self, lab_frame_vars ):
-        """
-        Converts a list of laser curvature from the lab frame
-        to the boosted frame:
-        R' = R / ( 1 + beta0 )
-
-        Parameters
-        ----------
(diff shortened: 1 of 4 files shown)


fbpic/lpa_utils/boosted_frame.py
```diff
@@ -219,28 +219,6 @@ def wavenumber( self, lab_frame_vars ):
 
         return( boosted_frame_vars )
```
RemiLehe: The functions below are not needed anymore because of the new way in which the boosted frame is performed.

fbpic/lpa_utils/laser/antenna.py
```diff
@@ -208,25 +222,14 @@ def update_v( self, t ):
         t: float (seconds)
             The time at which to calculate the velocities
         """
```
RemiLehe: The code below is not needed anymore, because it is performed directly inside the `gaussian_profile` function.

PR #50, opened 2017-06-01
Added bunch initialization from numpy arrays

RemiLehe:
Currently, the user can inject a particle bunch in an FBPIC simulation by passing the particle coordinates in a text file or openPMD file. However, it would be convenient to also have the possibility to pass numpy arrays directly.

This pull request implements a function that takes numpy array as argument. 
In addition, this function can also be used inside the other `add_elec_bunch_...` functions, in order to make them shorter.

### DIFF
--- a/docs/source/api_reference/lpa_utilities.rst
+++ b/docs/source/api_reference/lpa_utilities.rst
@@ -21,7 +21,9 @@ Beam utilities
 
 .. autofunction:: fbpic.lpa_utils.bunch.add_elec_bunch_gaussian
 
-.. autofunction:: fbpic.lpa_utils.bunch.add_elec_bunch_file	  
+.. autofunction:: fbpic.lpa_utils.bunch.add_elec_bunch_from_arrays
+
+.. autofunction:: fbpic.lpa_utils.bunch.add_elec_bunch_file
 
 External fields
 ---------------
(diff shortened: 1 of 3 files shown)


docs/source/api_reference/lpa_utilities.rst
```diff
@@ -21,7 +21,9 @@ Beam utilities
 
 .. autofunction:: fbpic.lpa_utils.bunch.add_elec_bunch_gaussian
```
RemiLehe: This is in order to have this function added in the FBPIC automatic documentation.

fbpic/lpa_utils/bunch.py
```diff
@@ -405,8 +387,8 @@ def add_elec_bunch_openPMD( sim, ts_path, z_off=0., species=None, select=None,
 
     # Transform particle distribution in
     # the Lorentz boosted frame, if gamma_boost != 1.
```
RemiLehe: I also corrected the lines below: it seems to be a previous error, since the method `boost_particles` does not return anything.

fbpic/lpa_utils/bunch.py
```diff
@@ -416,10 +398,10 @@ def add_elec_bunch_openPMD( sim, ts_path, z_off=0., species=None, select=None,
     # to allow for energy spread
     gamma0 = 1. / np.mean(relat_elec.inv_gamma)
     get_space_charge_fields( sim.fld, [relat_elec], gamma0,
```
RemiLehe: The changes below are not significant. Sorry for the distraction.

PR #53, opened 2017-06-08
Modify moving window and guard regions

RemiLehe:
**Please wait for milestone 0.4.0 (full ionization support) to be merged and released before merging this PR.**

New implementation of the guard regions and the moving window:

- additional variable "n_damp" introduced that defines the damping region at the left and right side of the box in case of "open" boundaries

- The guard region defined by "n_guard" is only for the MPI exchange of the fields and particles, and for removing / adding particles

- The "exchange_period", the number of timesteps after which the particles are exchanged / removed / added, is now calculated automatically in the simulation class. It is also no attribute of the boundary communicator anymore. (This will change again in a future PR)

- The "guard_cell_damper" class has been removed and its functionality is now integrated into the "boundary_communicator"

- It is not possible anymore to launch a parallel simulation with an infinite order stencil (n_order = -1) , as this would give wrong results

- Launching a parallel simulation with n_order=None, will result in having n_guard = 2*n_order (automatically set by the code). (In a future PR, the exact number of guard cells that are required will be calculated automatically)

- The injection of new particles (in the case of continous_injection) was modified. The number of new particles that are generated now depends only on "exchange_period" N_new_particles = (v_moving_window-v_plasma)*dt*exchange_period

- The moving window is now updated every timestep and the grids are also shifted every timestep. This is more robust than before. However, performance is negatively affected as the fields need to be copied / shifted every timestep and also rho_prev now is calculated every timestep. (Performance enhancements will be addressed in a future PR)


RemiLehe:
@MKirchen : Is this ready for review? If yes, could you update the description of the PR above (esp. for future reference) ? Thx ✨

fbpic/boundaries/boundary_communicator.py
```diff
@@ -198,6 +243,9 @@ def divide_into_domain( self, zmin, zmax, p_zmin, p_zmax ):
         self.iz_start_procs = [ k*Nz_per_proc for k in range(self.size) ]
         # Get the enlarged number of cells in each domain (includes guards)
         self.Nz_enlarged_procs = [ n+2*self.n_guard for n in Nz_domain_procs ]
+        # Add damping region to first and last domain
+        self.Nz_enlarged_procs[0] += self.n_damp
+        self.Nz_enlarged_procs[-1] += self.n_damp
```
RemiLehe: Note: at some point, we should check that this does not introduce load imbalance between the GPUs, i.e. maybe we could determine Nz_per_proc once we added the damp cells.

docs/source/example_input/boosted_frame_script.py
```diff
@@ -40,13 +40,18 @@
...
+
 # The simulation timestep
 dt = (zmax-zmin)/Nz/c   # Timestep (seconds)
 N_step = 101     # Number of iterations to perform
                  # (increase this number for a real simulation)
 
+# Stencil order of the solver. -1 corresponds to the standard
+# fully spectral solver with infinite order. For a localized stencil,
+# thus finite-order accuracy of the solver, choose n_order > 4
+# (needs to be a multiple of 2).
```
RemiLehe: Note: the above text should probably be rectified so as to fit within an 80 character line.

fbpic/boundaries/boundary_communicator.py
```diff
@@ -198,6 +243,9 @@ def divide_into_domain( self, zmin, zmax, p_zmin, p_zmax ):
         self.iz_start_procs = [ k*Nz_per_proc for k in range(self.size) ]
         # Get the enlarged number of cells in each domain (includes guards)
         self.Nz_enlarged_procs = [ n+2*self.n_guard for n in Nz_domain_procs ]
```
RemiLehe: This should only be done if boundaries are open, no?

fbpic/boundaries/boundary_communicator.py
```diff
@@ -208,16 +256,32 @@ def divide_into_domain( self, zmin, zmax, p_zmin, p_zmax ):
                                than 4 times n_guard. Use fewer domains or \
                                a smaller number of guard cells.')
 
-        # Calculate the local boundaries (zmin and zmax)
+        # Calculate the local boundaries (zmin)
         # of this local simulation box including the guard cells.
         iz_start = self.iz_start_procs[self.rank]
         zmin_local = zmin + (iz_start - self.n_guard)*dz
```
RemiLehe: This whole section of code could probably be simplified by doing:
```python
zmin_local_domain = zmin + iz_start*dz
zmax_local_domain = zmin_local_domain + self.Nz_domain
p_zmin = max( zmin_local_domain + self.n_guard*dz, p_zmin)
p_zmax = min( zmax_local_domain - self.n_guard*dz, p_zmax)
```
where `zmin_local_domain` and `zmax_local_domain` are the physical positions of the local **physical domain** (i.e. without guard and damp cells).

fbpic/boundaries/boundary_communicator.py
```diff
@@ -532,7 +680,7 @@ def gather_grid( self, grid, root = 0):
         if self.rank == root:
             # Calculate global edges of the simulation box on root process
             zmin_global = grid.zmin + self.dz * \
-                            (self.n_guard - self.rank*self.Nz_domain)
+                (self.n_guard + self.n_damp - self.rank*self.Nz_domain)
```
RemiLehe: I would still calculate more explicitly: 
```python
nz_invalid = self.n_guard
if self.left_proc is None:
    nz_invalid += self.n_damp
zmin_global = grid.zmin + self.dz * (nz_invalid - self.rank*self.Nz_domain)
```
Mostly because root could be different than 0.

fbpic/boundaries/boundary_communicator.py
```diff
@@ -583,7 +731,18 @@ def gather_grid_array(self, array, root = 0):
             gathered_array = None
         # Shortcut for the guard cells
```
RemiLehe: For more conciseness, I would do 
```python
n_left = self.n_guard
n_right = self.n_guard
# Remove n_damp cells from the left proc's output region
if self.left_proc is None:
    n_left += self.n_damp
# Remove n_damp cells from the right proc's output region
if self.right_proc is None:
   n_right += self.n_damp
```
Then the variable `ng` is not needed anymore.

fbpic/boundaries/moving_window.py
```diff
@@ -33,6 +33,14 @@ def __init__( self, interp, comm, ptcl, v, p_nz, time,
             Contains information about the MPI decomposition
             and about the longitudinal boundaries
 
+        exchange_period: int
```
RemiLehe: Shouldn't we remove ` after which the moving window is updated`, since the moving window is now updated at every single timestep ?

fbpic/boundaries/moving_window.py
```diff
@@ -90,8 +97,12 @@ def __init__( self, interp, comm, ptcl, v, p_nz, time,
 
         # Attach injection position and speed (only for the last proc)
         if comm.rank == comm.size-1:
+            self.v_end_plasma = \
+                c * self.uz_m / np.sqrt(1 + ux_m**2 + uy_m**2 + self.uz_m**2)
             ng = comm.n_guard
-            self.z_inject = interp[0].zmax - ng/2*interp[0].dz
+            nd = comm.n_damp
```
RemiLehe: Maybe we should add a comment here, like:
```
# Initialize plasma *ahead* of the right *physical* boundary of the box
# so, after `exchange_period` iterations (without adding new plasma),
# there will still be plasma inside the physical domain.
```

RemiLehe:
Off-line discussion: the requested modifications will be directly implemented in PR #55

PR #60, opened 2017-06-13
Changed cuda imports

MKirchen:
This simplifies the import of cuda related methods. CUDA availability is now always checked in cuda_utils and stored in the variable cuda_installed. This variable is imported at the beginning of all other cuda-related files.


RemiLehe:
Thanks! That simplifies the code a lot!!

RemiLehe:
Note: the tests pass on GPU with this version.

PR #64, opened 2017-06-15
Update the diagnostics

RemiLehe:
This pull request fixes a number of issues with the diagnostics:
 - The checkpoints are now done at the end of the PIC loop
 - `rho` and `J` are only brought back from spectral space if they are needed for the diagnostics
 - The code does not crash anymore when trying to write into a diagnostic file that previously existed (this issue was typically due to a different number of particles in the pre-existing diagnostic)

In addition, I updated and cleaned the comments in the main step function.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -315,17 +315,17 @@ class Simulation(object):
             if show_progress and self.comm.rank==0:
                 progression_bar( i_step, N, measured_start )
 
+            # Run the diagnostics
+            # (E, B, rho, x are defined at time n; J, p at time n-1/2)
+            for diag in self.diags:
+                # Check if the diagnostic should be written at this iteration
+                # and write it, if it is the case.
+                # (If needed: bring rho/J from spectral space, where they 
+                # were smoothed/corrected, and copy the data from the GPU.)
+                diag.write( self.iteration )
+
             # Exchanges to prepare for this iteration
             # ---------------------------------------
-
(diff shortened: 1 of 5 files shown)


fbpic/main.py
```diff
@@ -315,44 +319,46 @@ def step(self, N=1, correct_currents=True,
...
-            # (J, x, p are defined at time n+1/2)
+            # (E, B, rho, x are defined at time n; J, p at time n-1/2)
             for diag in self.diags:
-                # Check if the fields should be written at
-                # this iteration and do it if needed.
-                # (Send the data to the GPU if needed.)
+                # Check if the diagnostic should be written at this iteration
+                # and write it, if it is the case.
+                # (If needed: bring rho/J from spectral space, where they
+                # were smoothed/corrected, and copy the data from the GPU.)
```
MKirchen: Maybe we should add a note here, that says that E, B, rho, x are defined at n and J,p are defined at n-1/2

fbpic/main.py
```diff
@@ -259,8 +259,9 @@ def dens_func( z, r ) ...
 
         # Initialize an empty list of external fields
         self.external_fields = []
-        # Initialize an empty list of diagnostics
+        # Initialize an empty list of diagnostics and checkpoints
```
MKirchen: Maybe say also here, that checkpoints can be used for restarting a simulation

PR #55, opened 2017-06-10
Automatic calculation of n_guard based on the stencil reach for a given n_order

MKirchen:
**Please merge PR #53 first!**

This PR allows for automatic calculation of the guard regions size depending on the chosen order of the PSATD solver (for parallel computation).

- Adds the automatic calculation of n_guard for a given n_order depending on the stencil reach in spatial space.

- Clean-up of fields.py. Added a new file utility_methods.

- exchange_period is now calculated inside the boundary communicator again.

PR #65, opened 2017-06-15
Shift the fields in spectral space and update the diagnostic

RemiLehe:
**Please merge #55 first**

This pull request aims to improve performance by:
- **Shifting the fields (for the moving window) in spectral space:** This is because shifting the fields in real space requires a copy of arrays (in order to avoid race conditions) while shifting the fields in spectral space is only an in-place multiplication by a correct phase.
- **Avoiding to deposit `rho_prev` at each iteration:** Instead `rho_prev` is obtained directly from the previous iteration, **in spectral space**.  `rho_prev` is only brought back from spectral space to real space if it is needed for the diagnostics.

In addition, the diagnostics and checkpoints were updated (see PR #64 )


fbpic/boundaries/moving_window.py
```diff
@@ -412,3 +544,39 @@ def shift_field_array_gpu( field_array, field_buffer, n_move ):
...
+            this function
+
+        shift_factor: 1darray of complexs
+            Contains the shift array, that is multiplied to the fields in
+            spectral space to shift them by one cell in spatial space
+            ( exp(i*kz_true*dz) )
+
+        n_move: int
+            The number of cells by which the grid should be shifted
+        """
```
RemiLehe: Could we actually use `iz` and `ir` here, instead of `i` and `j` ? I think that would be clearer.
(The same would apply to the above kernel.)

fbpic/boundaries/moving_window.py
```diff
@@ -412,3 +544,39 @@ def shift_field_array_gpu( field_array, field_buffer, n_move ):
...
+        # Get a 2D CUDA grid
+        i, j = cuda.grid(2)
+
+        # Only access values that are actually in the array
+        if j < field_array.shape[1] and i < field_array.shape[0]:
+
+            if n_move > 0:
+                # Shift fields backward
+                field_array[i, j] = field_array[i, j] \
+                    * ( shift_factor[i] )**np.abs(n_move)
```
RemiLehe: No need for the `np.abs`. Then the function will be correct if `n_move` is negative.

fbpic/boundaries/moving_window.py
```diff
@@ -412,3 +544,39 @@ def shift_field_array_gpu( field_array, field_buffer, n_move ):
...
+            if n_move < 0:
+                # Shift fields forward
+                field_array[i, j] = field_array[i, j] \
+                    * ( -1.*shift_factor[i] )**np.abs(n_move)
```
RemiLehe: Same thing: no need for the `abs`.

fbpic/boundaries/moving_window.py
MKirchen: good catch! forgot to change that!

PR #73, opened 2017-06-19
Add test of charge conservation

RemiLehe:
This pull request adds a test of the charge conservation to the periodic plasma wave.

Because the simulation is periodic and we never cut the stencil, charge conservation should be satisfied to a precision close to the machine precision (when the current correction is used).


MKirchen:
Would you be fine with using order = 16 by default? That way we can always test the finite order part of the code and we can directly run the test in parallel also without changing the order to 16 by hand.

RemiLehe:
Sure, I will modify the PR accordingly.

PR #74, opened 2017-06-19
Fix error with cuda power operator

RemiLehe:
When running the tests on the GPU, I realized that the new function `shift_spect_array_gpu` is causing an error when running on GPU: the power operator (`**n_move`) does not seem to work correctly.

I by-passed this strange error by simply raising to the power `n_move` by hand (since `n_move` is typically 1 or otherwise a small integer).

### DIFF
--- a/fbpic/boundaries/moving_window.py
+++ b/fbpic/boundaries/moving_window.py
@@ -571,6 +571,9 @@ if cuda_installed:
 
         # Only access values that are actually in the array
         if ir < field_array.shape[1] and iz < field_array.shape[0]:
-            # Shift fields backward
-            field_array[iz, ir] = field_array[iz, ir] \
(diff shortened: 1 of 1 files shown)


RemiLehe:
I checked: all the automated tests pass on GPU for this branch.

MKirchen:
Thanks for finding this bug!

Maybe it would be nicer to use math.pow() (on the numba docs it says that it is supported with CUDA). However, I am also fine with what you proposed.

RemiLehe:
Yeah, I thought about `math.pow`. But then I read [here](https://docs.python.org/3/library/math.html#math.pow) that `math.pow` always converts the arguments to float, and then exponentiates the result. This probably very slow compared to the current implementation, in the case where `n_move` is an integer (which is most of the time equal to 1).

Issue #58, opened 2017-06-13
Check consistency of the openPMD files in the ionization test


RemiLehe:
Right now the ionization test checks the number of ionized ions from the particle arrays directly. It would be good to check that the number of ionized ions is consistent in the openPMD file.

PR #76, opened 2017-06-21
Fix and test ionization output in the boosted-frame

RemiLehe:
When implementing the automated tests for the openPMD output with ionization (see issue #58), I realized that there was a bug in the boosted-frame output for ionization (which caused the code to crash when using it!).

The present pull request fixes this bug and implements corresponding automated tests.

Note that this pull request is to the `master` branch and not the `dev` branch. This is intentional: once the pull request is merged, I'll do a new release (`0.4.1`) that corrects this bug. I will then merge `master` into `dev` so that `dev` also has the bug fix.


MKirchen:
Good that you found the bug. I'm okay with merging once the latest tests passed.

PR #78, opened 2017-07-03
Changes precision of directory naming in parametric example script

soerenjalas:
There is a problem, when using the parmetric_script.py, if you try to start a scan with closely spaced parameters, e.g. something like:

```python
a0_list = np.linspace(1.8, 2.2, 8)
```
i.e. values that require more than single digit precision.

For the diags the script creates directories with %.1f string replacements, meaning that 1.81 and 1.84 would create the same directory, which causes problems when writing the output.

As I couldn't think of a real robust solution I just changed the precision to %.2f. This doesn't fix the problem, but makes it less likely.

### DIFF
--- a/docs/source/example_input/parametric_script.py
+++ b/docs/source/example_input/parametric_script.py
@@ -135,7 +135,7 @@ if __name__ == '__main__':
     # Add a field diagnostic
     # Parametric scan: each MPI rank should output its data to a
     # different directory
-    write_dir = 'diags_a0_%.1f' %a0
+    write_dir = 'diags_a0_%.2f' %a0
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for this PR :)
I think that the tests are failing because you also need to substitute `.1f` for `.2f` in the `test_example_docs_scripts.py` file. This is because this automated test reads again the openPMD files to make sure that they have the right a0.
Could you make that change and update the PR? Thanks ✨

soerenjalas:
Thanks, that did it. If you like the change you can merge.

RemiLehe:
Great! Thanks for this!

PR #79, opened 2017-07-06
Change order of particles at initialization

RemiLehe:
Before this pull request, the particles were initialized such that, when going through the array of particles in order:
- `theta` changes the fastest (while `z` and `r` remain constant)
- then `z` changes 
- then `r` changes the slowest

However, in order to optimize cache reuse, in the current deposition and gathering, it is better to have the following order:
- `theta` changes the fastest (while `z` and `r` remain constant)
- then `r` changes 
- then `z` changes the slowest

This pull request implements the above change. 

Although this order is theoretically supposed to be more efficient (on CPU), I did not observe any significant speedup in practice. I think it still makes sense to merge this PR, as this is the most logical order for the particles to be in.

### DIFF
--- a/fbpic/particles/particles.py
+++ b/fbpic/particles/particles.py
@@ -182,9 +182,10 @@ class Particles(object) :
             # Get the corresponding particles positions
             # (copy=True is important here, since it allows to
             # change the angles individually)
-            zp, rp, thetap = np.meshgrid( z_reg, r_reg, theta_reg, copy=True)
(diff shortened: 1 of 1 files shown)


MKirchen:
Does this PR have any impact on the GPU performance?

RemiLehe:
In theory: no, since the particles are reordered anyway on the GPU.

And in practice: I tested the performance on the GPU (2000x400 box full of electrons) and saw no difference between the `dev` branch and this one.

PR #83, opened 2017-07-17
Implement parallel reduce

RemiLehe:
This pull request implements parallel reduction after the current deposition. It speeds up the reduction by roughly a factor of 2.

Here are the timings **before** this pull request (see green curve "reduce"
screen shot 2017-07-15 at 6 33 27 pm
)
and here is the timing **after** this pull request
screen shot 2017-07-16 at 11 33 59 pm

### DIFF
--- a/fbpic/particles/deposition/threading_methods.py
+++ b/fbpic/particles/deposition/threading_methods.py
@@ -12,7 +12,7 @@ import math
 from scipy.constants import c
 
 # -------------------------------
-# Particle shape Factor functions 
+# Particle shape Factor functions
 # -------------------------------
 
 # Linear shapes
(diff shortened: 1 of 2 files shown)


fbpic/particles/deposition/threading_methods.py
```diff
@@ -427,3 +427,34 @@ def deposit_J_prange_linear(x, y, z, w,
...
+    """
+    # Extract size of each dimension
+    Nreduce, Nz, Nr = global_array.shape
+
+    # Parallel loop over iz
+    for iz in prange( Nz ):
+        # Loop over the reduction dimension (slow dimension)
+        for it in range( Nreduce ):
+            # Loop over ir (fast dimension)
+            for ir in range( Nr ):
```
MKirchen: Wouldn't it be possible to use prange for Nr as well? In that case there might be some vectorisation possible?

PR #96, opened 2017-07-22
Remove calls to the old gather functions and deposit functions

RemiLehe:
This changes `particles.py` so as to remove the calls to the old gathering and deposit functions (which used the `weights` function). 

Gathering and deposition is now done with the new threaded functions. When threading is disabled (`FBPIC_DISABLE_THREADING=1`), these functions still work, through the njit_parallel/prange trick. (Therefore I inserted `njit_parallel/prange` in all the threaded gathering and deposit functions.) 

Also, in order for the deposition to still work with the new threaded code, I had to define `self.nthreads=1` in the case where threading is disabled. This means that, even when we disable threading, we still create a temporary array (of size 1 along the threading dimension), and then we still do the `sum_reduce` (which basically a simple sum because the table is of size 1 along the threading dimension). We could probably save some time there, but in practice the `sum_reduce` is really a minor cost (and the whole operation is so much faster than when we were using `weights` anyway.)

Note that I did not remove the `weights` function, nor did I remove the code for the old deposition function. This is because they are used in the laser antenna (which would require more work to switch to the threaded functions). I did however remove the code for the old gathering functions.


RemiLehe:
Note: I ran the automated tests with `FBPIC_DISABLE_THREADING=1`. Everything works fine.

RemiLehe:
Note: I also explicitly set the number of threads in the Travis tests to 2. This suddenly made the tests much faster (from ~40 min to ~10 min total). Before this, Travis was probably trying to use too many threads, which lead to this very long testing time.

RemiLehe:
It seems that Travis recently switched a different default distribution ("trusty"): https://blog.travis-ci.com/2017-07-11-trusty-as-default-linux-is-coming.

However, for some reason the tests on "trusty" are extremely slow, and do not reach completion. I therefore set the distribution to "precise" (the previous default).

MKirchen:
I am not sure if we should already merge this.

1. we still need the old deposition method for the antenna, so we need to keep this code anyways for now
2. prange is very new and maybe there might occur bugs or api changes in the future. keepinh the old implementation as a reference might be a good idea at least for some time.
3. the old reference implementation with weights is maybe easier to understand for new developers? (maybe we should keep it therefore?)

Apart from that: I dont like that the gathering file is now called numba_methoda again, while for the deposition we have numba_methods and threading_methods. this is somehow confusing.

I dont know, I am not fully sure what we should do... You can decide wether to merge or not!

RemiLehe:
OK, I'll try to think about the points that you raised and I'll see if I can find a better way.

RemiLehe:
I thought again about this PR, and updated/modified slightly.

In particular, I renamed the file that contains the gathering methods, so that it is now again called `threading_methods.py`, as you suggested it would be more consistent.

Regarding the other points that you raised:

- It is true that `prange` is new, but it can be deactivated using `FBPIC_DISABLE_THREADING` (in which case it becomes a normal `range`). And in that case, the new functions are still faster than the old ones which used the `weights` functions. So I think that we can safely remove the calls to these old functions. If at some point we realized that we need them again, we can easily go back in the git history and copy them again.

- Regarding the fact that the old methods might be more understandable: I think it is a trade-off, since the code in `particles.py` is then also quite a bit longer. At least, with this PR, all the complicated/confusing part is isolated into a separate file (`threading_methods.py`).

Given the above, would you be okay to merge this? If you prefer to wait a bit more, that's also fine with me.

MKirchen:
okay!

PR #98, opened 2017-07-24
`w` should be the actual particle weight

RemiLehe:
In the code before this PR, the particle quantity `w` represented different things:
- For non-ionizable particles: it was the weight (i.e. number of physical particle per macroparticle) times the charge.
- For ionizable particles: it was the weight times the ionization level, times the elementary charge e. This also implied that we had to keep an additional quantity (`neutral_weight`) to store the actual number of physical particle per macroparticle, inside the `Ionizer` class.

This is confusing, and therefore the current pull request sets `w` to be the actual weight (i.e. number of physical particle per macroparticle) for both non-ionizable particles and ionizable particles. 
- For non-ionizable, this implies that `w` has to be multiplied by the charge inside the deposition kernel (this is a minor cost)
- For ionizable particles, `w` now plays the part that `neutral_weight` used to play. And therefore, inside the Ionizer class, `neutral_weight` is replaced by a new array `w_times_level`, which is only used for the deposition kernel, as the effective weight.

Note that a lot of the changes involve replacing `neutral_weight` by `w_times_level` in all the MPI communications. The code for the MPI communications of these optional quantities (e.g. `w_times_level`, `ionization_level`, etc.) is quite ugly, and should at some point be replaced by something better.

PS: In fact, `w_times_level` does not need to be communicated via MPI, since it can be recalculated at each `exchange_period`, from `w` and `ionization_level`. I kept the MPI communications mostly to minimize the changes in the code that this PR produces.

### DIFF
--- a/fbpic/boundaries/particle_buffer_handling.py
+++ b/fbpic/boundaries/particle_buffer_handling.py
@@ -126,7 +126,7 @@ def remove_particles_cpu(species, fld, n_guard, left_proc, right_proc):
         if species.ionizer is not None:
             uint_send_left[i_attr,:] = \
                 species.ionizer.ionization_level[selec_left]
-            float_send_left[8,:] = species.ionizer.neutral_weight[selec_left]
+            float_send_left[8,:] = species.ionizer.w_times_level[selec_left]
     else:
         # No need to allocate and copy data ; return an empty array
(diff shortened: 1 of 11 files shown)


MKirchen:
Nice PR! Good changes!

RemiLehe:
Thanks for reviewing!

PR #100, opened 2017-07-26
Add threading to ionization routines

RemiLehe:
This pull request introduces threading in some of the used by the ionization routines.

In addition, I also bundled to small changes:
- A correction of a bug related to PR #98 (missing replacement of `neutral_weight` by `w_times_level`)
- An update of the install instructions on Lawrencium

### DIFF
--- a/docs/source/install/install_lawrencium.rst
+++ b/docs/source/install/install_lawrencium.rst
@@ -56,8 +56,11 @@ Installation of FBPIC and its dependencies
 
    ::
 
-       conda install -c conda-forge numba scipy h5py pyfftw mpi4py accelerate
-       conda install -c numba cudatoolkit=8.0
+       conda install numba spicy h5py
+       conda install -c conda-forge mpi4py
+       conda install -c numba cudatoolkit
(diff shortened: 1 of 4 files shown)


RemiLehe:
Here is how the timing for the ionization routine compares to the other routines on a 2000x400 grid filled with Helium that gets progressively ionized by a laser pulse.

screen shot 2017-07-27 at 3 23 13 pm

RemiLehe:
And here is a breakdown of the different components of the ionization routine:
screen shot 2017-07-27 at 3 23 05 pm

RemiLehe:
A few comments on the scalings:
- random generator: this is purely serial so far, so no surprise that it does not depend on nthreads.
- handle_ionization_cpu: this is the high level function that calls the other. I am guessing that the time taken in this function (itself) is essentially array allocation and Python overhead.
- ionize_ions_numba: good scaling! Not so suprising, as it is a compute-intensive part.
- copy_particle_data_numba: small function that we call many times (once per particle quantity). So the saturation with nthreads is probably essentially function call overheads.

MKirchen:
:)

PR #106, opened 2017-07-31
Generalized ionization

RemiLehe:
This pull request updates the ionization code so as to:
- reduce duplications between the CPU and GPU code
- generalize the code so that similar types of processes (where some macroparticles create other macroparticles) are easier to implement and can reuse some of the code that was written for the ionization.

By "similar types of processes", I am mainly thinking about Compton scattering (which I implemented on a separate branch, and which uses the changes of the this PR).

In order to generalize the code, I created a new folder `elementary_process` in the `particles` folder, and moved the `ionization` folder into `elementary_process`.  (on the `compton_scattering` branch, the `elementary_process` branch also contains a folder `compton`.) All the functions which were general enough to be reused by other elementary processes were moved from the `ionization` folder to the `elementary_process` folder.


RemiLehe:
According to the Numba developers: there should be no difference in performance, with the current version of Numba
https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/SMcDsfNlP1k

fbpic/main.py
```diff
@@ -368,11 +368,6 @@ def step(self, N=1, correct_currents=True,
             for ext_field in self.external_fields:
                 ext_field.apply_expression( self.ptcl, self.time )
```
RemiLehe: The modifications below are a correction of a minor mistake: the particles should have their position and momenta synchronized (i.e. at the same timestep), in order to calculate the ionization.

fbpic/particles/elementary_process/cuda_numba_utils.py
```diff
@@ -0,0 +1,130 @@
+# Copyright 2017, FBPIC contributors
```
RemiLehe: This is the generic file that contains methods that can be used by ionization, but also other elementary processes (e.g. Compton)

fbpic/particles/elementary_process/ionization/cuda_methods.py
```diff
@@ -9,8 +9,16 @@
 """
 from numba import cuda
 from scipy.constants import c
-from .inline_functions import get_ionization_probability_cuda, \
-    get_E_amplitude_cuda, copy_ionized_electrons_batch_cuda
```
RemiLehe: This is a minor change:

The inline functions used to be compiled inside the file `inline_functions`, for both GPU and CPU. This imposed that we give different names to those compiled functions. 

Now they are compiled for GPU in the file `cuda_methods.py` and for CPU in the file `numba_methods.py`. Thus we can keep the same name of the compiled function on GPU and CPU, which reduces divergence between the GPU and CPU code.

fbpic/particles/elementary_process/ionization/inline_functions.py
```diff
@@ -31,11 +26,6 @@ def get_E_amplitude( ux, uy, uz, Ex, Ey, Ez, cBx, cBy, cBz ):
 
     return( math.sqrt( E2_on_particle ), gamma )
```
RemiLehe: Remove compilation (compilation is now done in the files that import these inline functions.)

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -156,9 +157,10 @@ def initialize_ADK_parameters( self, element, dt ):
         if self.use_cuda:
             self.prng = PRNG()
```
RemiLehe: The changes below correspond to the reduction of the duplication between the CPU and GPU code. Instead of having two methods (`handle_ionization_gpu` and `handle_ionization_cpu`), I now have a single method (`handle_ionization`), which uses code relevant to GPU or CPU only when needed.

fbpic/particles/particles.py
```diff
@@ -371,15 +371,13 @@ def make_ionizable( self, element, target_species,
...
-        Ionize this species, and add new macroparticles to the target species
+        Handle elementary processes for this species (e.g. ionization)
         """
+        # Ionization
         if self.ionizer is not None:
-            if self.use_cuda:
-                self.ionizer.handle_ionization_gpu( self )
-            else:
-                self.ionizer.handle_ionization_cpu( self )
+            self.ionizer.handle_ionization( self )
```
RemiLehe: In the `compton_scattering` branch, there is another function call here, in order to handle Compton scattering.

fbpic/particles/particles.py
```diff
@@ -434,6 +432,10 @@ def push_p( self ) :
         half-timestep *behind* the positions (x, y, z), and it brings
         them one half-timestep *ahead* of the positions.
         """
```
RemiLehe: The changes below allow to have photon macroparticles in the code (q=0): these macroparticles should not deposit charge, nor should they have their momentum modified or gather the fields.

PR #109, opened 2017-08-04
Remove function signatures on GPU

RemiLehe:
This pull request removes all the function signatures for the cuda methods.

The main aim of this is to reduce the compilation time at the beginning of the simulation: I recently realized that all the functions that have signatures get compiled **when they are imported**. By contrast, the functions that do not have signatures get compiled **when they are first called**.

This means that, in the current dev branch, the deposition and gathering function for **both 1st order and 3rd order** get compiled before the simulation even starts, even though we usually only use one type of shape factor. The corresponding compilation time slows down a lot the startup time for an FBPIC GPU simulation. 

The aim of this pull request is that only the GPU functions that we do need get compiled. 

I did a few tests with a 2000x400 box filled with plasma, performing 10 steps first (this includes just-in-time compilation) and then 100 steps (without compilation, since all functions have already been compiled).

- with the current `dev` branch, here is the output:
```
Running FBPIC on GPU with 1 proc.
[--------------------------------------------------] 9/10, 0:00:00 left
 Time taken by the loop: 0:00:12
[--------------------------------------------------] 99/100, 0:00:00 left
 Time taken by the loop: 0:00:48
```
but the total time taken by the test is `0:03:30`! (measured separately) This is very likely due to the compilation even before the first 10 steps start.

- with this pull request:
```
Running FBPIC on GPU with 1 proc.
[--------------------------------------------------] 9/10, 0:00:00 left
 Time taken by the loop: 0:00:32
[--------------------------------------------------] 99/100, 0:00:00 left
 Time taken by the loop: 0:00:48
```
and the total time is `0:01:52`.

Conclusions:
- The first 10 steps takes more time because **all** the required functions get compiled during the first step.
- But overall, we save more than 1min30 (`0:03:30` vs `0:01:52`), very likely because we do not compile unneeded functions anymore.
- The time taken by the next 100 steps does not depend on whether the functions signatures are present or not.

### DIFF
--- a/fbpic/boundaries/cuda_methods.py
+++ b/fbpic/boundaries/cuda_methods.py
@@ -7,12 +7,7 @@ It defines a set of generic functions that operate on a GPU.
 """
 from numba import cuda
 
-@cuda.jit('void( complex128[:,:,:], complex128[:,:,:], \
-                 complex128[:,:], complex128[:,:], complex128[:,:], \
-                 complex128[:,:], complex128[:,:], complex128[:,:], \
-                 complex128[:,:], complex128[:,:], complex128[:,:], \
-                 complex128[:,:], complex128[:,:], complex128[:,:], \
-                 int32, int32, int32 )')
+@cuda.jit
 def copy_EB_to_gpu_buffers( EB_left, EB_right,
                             Er0, Et0, Ez0, Br0, Bt0, Bz0,
                             Er1, Et1, Ez1, Br1, Bt1, Bz1,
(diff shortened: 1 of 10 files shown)


RemiLehe:
For the record, here is the code that I used, in order to automatically remove the signatures:
```python
import glob
import re

for filename in glob.iglob('fbpic/**/*.py', recursive=True):
    print(filename)
    with open( filename ) as f:
        script = f.read()
    script =  re.sub(r"cuda.jit\('[\s\S]+?'\)", r"cuda.jit", script)
    match = re.search(r"cuda.jit\('[\s\S]+?'\)", script)
    assert match is None
    with open( filename, 'w' ) as f:
        f.write( script )
```

MKirchen:
I am happy to merge this if we are really sure that it does not affect performance?

RemiLehe:
You are right: we should be sure about this point. I'll ask on the Numba discussion group if there is a difference in overhead between functions with signatures and without.

RemiLehe:
According to the Numba developers: there should be no difference in performance, with the current version of Numba:
https://groups.google.com/a/continuum.io/forum/#!topic/numba-users/SMcDsfNlP1k

PR #112, opened 2017-08-09
Improve scaling of 3rd order gathering with number of threads

RemiLehe:
This pull request allows the 3rd order gathering to scale much better with the number of threads by avoiding the creation **for each particle** of small-sized arrays (that contain e.g. the particle shape).

Instead, these arrays are created only once for each thread (and are private to this thread), and then each thread loops over the particles in its chunk.

This uses a very similar chunking as for the deposition functions. As a consequence, I created a helper function in `threading_utils.py` which returns the array of chunk indices. (used both for the deposition, and for the 3rd gathering)

Here is the scaling with the current PR:
screen shot 2017-08-09 at 4 30 04 pm
By contrast, here is the scaling with the current `dev` branch:
current_dev
(Please disregard the curve for the Hankel transform on this last one.)

### DIFF
--- a/fbpic/particles/gathering/threading_methods.py
+++ b/fbpic/particles/gathering/threading_methods.py
@@ -384,14 +384,15 @@ def gather_field_numba_cubic(x, y, z,
         The magnetic fields acting on the particles
         (is modified by this function)
     """
-    ir = np.zeros( (nthreads, 4), dtype=int64) #[0,0,0,0]
-    Sr = np.zeros( (nthreads, 4) )
-    iz = np.zeros( (nthreads, 4), dtype=int64) #[0,0,0,0]
-    Sz = np.zeros( (nthreads, 4) )
-    
     # Gather the field per cell in parallel
     # (for threads < number of particles)
     for nt in prange( nthreads ):
+
+        ir = np.empty( 4, dtype=int64)
+        Sr = np.empty( 4 )
+        iz = np.empty( 4, dtype=int64)
+        Sz = np.empty( 4 )
+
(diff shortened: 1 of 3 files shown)


fbpic/particles/gathering/threading_methods.py
```diff
@@ -382,220 +383,237 @@ def gather_field_numba_cubic(x, y, z,
...
+        # to store the particle index and shape
+        ir = np.empty( 4, dtype=int64)
+        Sr = np.empty( 4 )
+        iz = np.empty( 4, dtype=int64)
+        Sz = np.empty( 4 )
+
+        # Loop over all particles in thread chunk
+        for i in range( ptcl_chunk_indices[nt],
+                            ptcl_chunk_indices[nt+1] ):
+
```
RemiLehe: The only change to this function is the above modification to the loop.  (Looping over nthreads and then over particles within each chunk, instead of looping directly over particles.)

The rest of this function is unchanged: it is simply indented to the right because of the new nested loop. (the only exception is the `np.zeros` calls which were removed from the body of the function below)

fbpic/particles/particles.py
```diff
@@ -220,10 +220,9 @@ def dens_func( z, r ) ...
                                         dtype=np.int32 )
             # Register boolean that records if the particles are sorted or not
             self.sorted = False
-        # Register variables when using multithreading
-        self.use_threading = threading_enabled
+
         # Register number of threads
```
RemiLehe: Note: I also removed `self.use_threading` in the `Particles` class, as it is not used anymore (since the PR in which we removed the old methods using the`weight` function)

PR #113, opened 2017-08-11
[WIP] Compton scattering

RemiLehe:
**Do not merge; this is work in progress**

This branch is functional, but:
- [ ] I need to write the docstrings (esp. the docstring should emphasize the conditions of validity of the Compton model: low a0 ; does not take into account the spectral broadening from the final length of the pulse ; does not produce photons at the harmonics of the resonant frequency)
- [ ] The API may not be stable at that point
- [ ] It would be nice to add the option to take polarization into account


RemiLehe:
A new, separate branch was created, where the current `dev` was merged with the Compton branch and conflicts were resolved. This branch is kept for potential comparisons in the future.

PR #114, opened 2017-08-28
MPI field exchange refactoring

MKirchen:
This PR includes a refactoring of the MPI exchange routines for the fields E,B as well as the source terms J and rho. In main.py, changes are included such that only J is added each timestep after the current correction, while rho is not exchanged any more.

These changes are inline with the upcoming changes of the current-correction (cross-depositon) for that are required for the new MPI release of FBPIC.

Note: The changes for including the new current correction have previously been included in this branch, but have been removed for the PR in order to have a cleaner PR.

Note: In field_buffer_handling.py, some lines are longer than 80 characters (PEP8), but I did not want to shorten them as this would reduce readability significantly. @RemiLehe What do you think?


fbpic/boundaries/boundary_communicator.py
```diff
@@ -192,9 +192,11 @@ def __init__( self, Nz, zmin, zmax, Nr, rmax, Nm, dt,
...
             cells_per_step = 2.*c*dt/self.dz
             # Maximum number of timesteps before a particle can reach the end
-            # of the guard region including the maximum number of cells (+/-3)
-            # it can affect with a "cubic" particle shape_factor.
-            self.exchange_period = int( (self.n_guard-3)/cells_per_step )
+            # of the half of guard region including the maximum number of cells
+            # (+/-3) it can affect with a "cubic" particle shape_factor.
+            # (Particles are only allowed to reside in half of the guard
+            # region as this is the stencil reach of the current correction)
+            self.exchange_period = int(((self.n_guard/2)-3)/cells_per_step)
```
MKirchen: Although this is not required to be included in this PR, I left it here so that we don't forget about it.

RemiLehe:
Thanks for this PR!
Before I review it, could you quickly check why the automated tests are failing?

Thanks!

fbpic/boundaries/boundary_communicator.py
MKirchen: With the newest change in nz_inject (+3 cells filled with plasma to the right), we need to make sure that "exchange_period" is small enough in the case of a boosted frame simulation with streaming plasma, such that particles can never be initialized in a region greater than n_guard/2. (this is not important for this PR, but only for the PR of the cross_deposition)

@RemiLehe When we implement the cross_deposition, i think that no particles are allowed to be in the outer half of the guard_region (assuming that the stencil of the cross_depositon is n_guard/2), right?

RemiLehe: OK, thank. I'll review the PR over the weekend, and I'll try to answer the above question.

fbpic/boundaries/field_buffer_handling.py
```diff
@@ -31,301 +34,449 @@ def __init__( self, n_guard, Nr, Nm, left_proc, right_proc ):
...
-                    copy_left, copy_right, ng )
-                # Copy the GPU buffers to the sending CPU buffers
-                if copy_left:
-                    self.d_EB_l.copy_to_host( self.EB_send_l )
-                if copy_right:
-                    self.d_EB_r.copy_to_host( self.EB_send_r )
+                if method == 'replace':
+                    # Copy the inner regions of the domain to the buffers
+                    copy_vec_to_gpu_buffer[ dim_grid_2d, dim_block_2d ](
+                        self.d_vec_rep_buffer_l, self.d_vec_rep_buffer_r,
```
RemiLehe: It seems that the variables `self.d_vec_rep_buffer_l` and `self.d_vec_rep_buffer_r` do not exist and cause the code to crash, when executed on GPU. 

@MKirchen : Could you try to run a test on GPU with MPI, and confirm that you also get an error?

RemiLehe:
I think we can merge this one too now.
I do have a few additional suggestions, but I'll make them via PR directly to the dev branch.

MKirchen:
as you like. you can merge or you can comment on the suggestions and i implement them.

PR #116, opened 2017-09-01
Close #115  Boosted particle diagnostic for GPU miss particles

### LINKED ISSUE #115: Boosted particle diagnostic for GPU miss particles
I recently observed that the boosted particle diagnostics fails to retrieve all particles. For instance, in the case of a Gaussian bunch with 30000 macroparticles, the boosted particle diagnostics write approx 25000 macroparticles only.

This issue on occurs on GPU ; on CPU, the full 30000 macroparticles are written.

RemiLehe:
This pull request fixes a bug with the boosted particle diagnostic (prefix_sum_shift was not taken intro account).

I also added an additional automated test, which checks that the boosted diagnostic capture all the particles that they should capture. With the current `dev` branch, this test does not pass on GPU, but it does pass on GPU with the changes from this PR.


MKirchen:
Thanks for the fix!

PR #119, opened 2017-09-08
Fix MPI GPU code

RemiLehe:
This pull request modifies the branch `mpi_refactor`, in such a way that the error that showed on GPU are corrected.

I also refactored some of the comments.


RemiLehe:
Note: I tested this on GPU, and it ran fine.

MKirchen:
Thanks Remi, I had a look at the code. You can merge whenever you want.

If you have more change requests for my PR, I can work on it tomorrow.

RemiLehe:
Ok thanks. I do have a few more suggestions, which I will also submit as PR to the mpi_refactor branch (as it is easier to explain this way)

PR #120, opened 2017-09-17
Correct MPI exchange pattern of J and rho

RemiLehe:
This pull request modifies the branch `mpi_refactor`:

- During the diagnostics, it takes into account the fact that `rho` is never exchanged during the PIC loop.
- It corrects the exchange of J in the case where correct_currents is False (In the previous version, J was not exchange in this case, whereas it should, for the purpose of pushing the Maxwell equations)
- It prevents the use of `use_true_rho` with MPI.

### DIFF
--- a/docs/source/example_input/boosted_frame_script.py
+++ b/docs/source/example_input/boosted_frame_script.py
@@ -189,5 +189,5 @@ if __name__ == '__main__':
                     ]
 
     ### Run the simulation
-    sim.step( N_step, use_true_rho=True )
+    sim.step( N_step )
     print('')
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -286,7 +286,7 @@ class Simulation(object):
             Whether to correct the divergence of E in spectral space
 
         use_true_rho: bool, optional
-            Wether to use the true rho deposited on the grid for the
+            Whether to use the true rho deposited on the grid for the
             field push or not. (requires initialize_ions = True)
 
         move_positions: bool, optional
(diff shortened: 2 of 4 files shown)


MKirchen:
Thanks for the fixes! Although the new PIC loop is not the most elegant solution (would have been nicer if we would not need to forbid use_true_rho & correct_divE with MPI), i can't think of any better solution at the moment that doesn't come with a lot more computational costs.

Still, after looking at the code, I have some more suggestions:

- Could you raise an error if someone tries to use correct_divE with MPI? (as you already mentioned on Slack)

- Could you also add an MPI exchange of rho after finalizing the loop. Thereby, at the end of the loop  rho and J are both consistent (i.e. filtered and exchanged).
```
        # Finalize PIC loop
        # Get the charge density and the current from spectral space.
        fld.spect2interp('J')
        fld.spect2interp('rho_prev')
        self.comm.exchange_fields(self.fld.interp, 'rho', 'add')
```

Another possible change, which might make the code a little bit more clean and easier to understand, would be the following...
```
            # Get the charge density at t = (n+1) dt
            self.deposit('rho_next')
            # Correct the currents (requires rho at t = (n+1) dt )
            if correct_currents:
                fld.correct_currents()
            # Exchange the (corrected) currents between MPI domains
            if self.comm.size > 1:
                fld.spect2interp('J')
                self.comm.exchange_fields(fld.interp, 'J', 'add')
                fld.interp2spect('J')
```
...while reverting again the changes that you did to the deposit function (i.e. NOT exchanging the current in this function).

This change would add computational costs in case one uses MPI without current correction (because of the additional spect2interp / interp2spect). However, in the future I think the most consistent way to use the code is with current correction always on. Therefore, I think we can accept that if we turn the current correction off, we have some small overhead.

You can decide what you prefer.

RemiLehe:
Thanks for your suggestions. I implemented the additional check with `div E` and the final exchange of `rho`. 

Regarding your suggestion on the exchange of J: I agree it would simplify a code significantly, which would be great. However, I think that the computational savings that we get in the case `correct_current=False` (i.e. saving a forward and a backward **vector** spectral transform) are not entirely negligible. Therefore, I would prefer (at least for now) to keep it, if it's okay.

MKirchen:
Okay!

PR #121, opened 2017-09-17
Avoid the notation [::-1]

RemiLehe:
This is more of a suggestion: should we replace the structure `[::-1][...][::-1]` by a single indexing operation (see the code) ?

The new code feels more concise and easier to understand, to me. But it is also a matter of personal preference.

@MKirchen : Feel free to reject this PR if you disagree.


RemiLehe:
Reopening this branch, after off-line discussion with @MKirchen

PR #123, opened 2017-09-20
Recomend numba 0.34 in documentation

RemiLehe:
With the current dev branch, if someone tries to install FBPIC according to the documentation, they will get runtime errors because they will have numba 0.35 by default, which fails with FBPIC when trying to use threading.

This PR corrects the documentation accordingly.


MKirchen:
Approved

RemiLehe:
Thanks :)

PR #125, opened 2017-09-23
Close #107 : Check that `add_elec` particles are within the local domain

### LINKED ISSUE #107: Functions in `lpa_utils/bunch.py` need to be fixed
The functions that create new electron bunches and calculate their space charge field have several issues that need to be fixed:
- [x] The initialization of the weight `w` has not been update with respect to the new definition of `w` (see #98)
- [ ] Nothing checks that the particles are in the local subdomain before adding them to the simulation. As a consequence, the subsequent particle deposition can lead to out-of-bounds (observed in practice).

RemiLehe:
The different versions of the function `add_elec_bunc` (e.g. Gaussian, uniform) were incorrect when used with MPI: in this case each proc initializes the full number of particles, and keeps them even if they are outside of the local physical domain.

This PR corrects this: each proc now retains only the particles that are within the local physical domain.

Note: in order to implement this, I needed a function that gives the zmin and zmax of the local domain. I implemented it as a method of the boundary communicator and used it in several other places in the code.

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -314,12 +314,11 @@ class BoundaryCommunicator(object):
 
         # Calculate the enlarged boundaries (i.e. including guard cells
         # and damp cells), which are passed to the fields object.
-        zmin_local_enlarged = zmin_local_domain - self.n_guard*dz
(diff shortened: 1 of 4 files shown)


MKirchen:
Nice :)

Issue #129, opened 2017-09-28
Optimize number of threads per block for each kernel


RemiLehe:
This could be done with a standard simulation script with moving window and box full of electrons.

Even if the optimized number of threads might not generalize to any other situation, I think it will still be better than the current arbitrary choice (i.e. 256 threads per block for all kernels)

RemiLehe:
That sounds great!! Thanks a lot for taking care of this ; I think we can really some non-negligible speed-up from this!

soerenjalas:
I did some preliminary test on some Nvidia K40, but the results are rather inconclusive.
I used a simulation with a 2000x300 cell grid filled with particles for 1000 steps. The benchmarks scripts are attached below and can be used with the branch above.
It seems that especially with the computational heavy functions like deposition and field push there is a lot to gain from optimising the threads per block. However, it seems quite random which tpb values work well. I did the test on DESYs Maxwell cluster and it could still be that some nodes perform better than others I will test this on Jureca.

![1dtpb](https://user-images.githubusercontent.com/10922951/31225453-747be344-a9d2-11e7-8406-e16ed5f7a591.png)
![2dtpb](https://user-images.githubusercontent.com/10922951/31225455-789563b0-a9d2-11e7-9b1c-df7a779840e1.png)


[benchmark_tpb1D.zip](https://github.com/fbpic/fbpic/files/1359125/benchmark_tpb1D.zip)
[benchmark_tpb2D.zip](https://github.com/fbpic/fbpic/files/1359126/benchmark_tpb2D.zip)

PR #131, opened 2017-10-10
Set threads per block of deposition and gathering kernels to 64

soerenjalas:
As discussed in issue #129 the run time of the GPU deposit and gathering kernels can be improved by reducing the number of threads per block.
This is the minimal effort fix by just hardcoding the number of threads in the `deposit` and `gather` method to 64. 
Note, that the benchmarks in #129 show the unexpected behaviour that `deposit_rho` performs best with 48 threads per block, this is ignored in this PR.


RemiLehe:
Thanks for the tests and the PR!

PR #132, opened 2017-10-21
Close #128: Documentation for the boosted frame

### LINKED ISSUE #128: Add documentation for boosted-frame
This could be added to the `Advanced` section.

We should probably also add a safe-guard that prevents the user from having c dt > dr in the boosted frame case (since this becomes unstable).

RemiLehe:
This introduces the boosted-frame technique in the documentation. 

You can get a preview of the documentation on this non-official link:
http://www.normalesup.org/~lehe/html_fbpic/advanced/boosted_frame.html


MKirchen:
Nice! Thanks for writing this!

I was wondering: Could we agree on setting use_galilean =True by default in the simulation object? I think its more consistent and we need to use it anyways as default when running the Galilean frame with finite stencils. If you have objections, I would at least mention in the documentation that one should set use_galilean = True when using the boosted frame. Also, I think it would be good to mention the exact formula for v_moving for a given gamma_boost. I just want to avoid that people use parameters that are not exactly matching to the NCI elimination (which need to be exact when running in parallel in the future), because at some point this might get confusing.

RemiLehe:
Thanks for reviewing this!
I definitely agree with switching to `use_galilean=True` (in fact, I thought we had made this change earlier, but apparently not).

I modified the code accordingly (see the last commit). I also added the exact v_comoving in the doc.

PR #133, opened 2017-10-24
Close #102 Experiment with MKL fft

### LINKED ISSUE #102: Experiment with MKL fft
We recently observed that the latest `pyfftw` is sometimes surprisingly slow on CPU. 
It would be good to try the MKL fft (which should normally be automatically called by `np.fft.fft`)

RemiLehe:
This pull requests adds the option to use the MKL FFT in FBPIC (note that the MKL library was usually automatically downloaded when installing the dependencies of FBPIC, even before this pull request). Because the performance tests are quite favorable (see below), the MKL FFT is used preferentially, and `pyfftw` is only used when MKL is not available. This also means that `pyfftw` is not, strictly speaking, a dependency of FBPIC anymore.

Regarding performance, here are the results of scaling with the number of threads, on a 24-core Intel node:

- with pyfftw:
screen shot 2017-10-23 at 10 45 52 pm

- with MKL FFT:
screen shot 2017-10-23 at 10 45 13 pm

The red curve (fft routines) has much lower compute cost, and also much better scaling with the number of threads in the case of MKL!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,8 +20,8 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes numba==0.34 scipy h5py
-  - conda install --yes -c conda-forge pyfftw mpi4py
+  - conda install --yes numba==0.34 scipy h5py mkl
+  - conda install --yes -c conda-forge mpi4py
   - pip install pyflakes
   - python setup.py install
 before_script:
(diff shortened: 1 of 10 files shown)


MKirchen:
Awsome! I agree to merge once the tests pass.

RemiLehe:
Thanks for reviewing this! Note: I also updated the documentation (installation section) to remove pyfftw.

PR #134, opened 2017-10-26
Add a section on how to profile the code with MPI

RemiLehe:
I modified the doc to explain how to profile MPI FBPIC simulation. This can be interesting in order to see load-imbalance, how much time is spent in communications, etc. The method is quite cumbersome, but it is the best one I could think of. If you have other suggestions, don't hesitate.

I also reorganised the rest of the profiling section, so as to group it primarily by CPU/GPU rather than text-based/GUI-based.

The rendered page can be seen on this (unofficial) link:
http://www.normalesup.org/~lehe/html_fbpic/advanced/profiling.html


MKirchen:
Very nice! This section is really helpful.

PR #136, opened 2017-11-03
Correct 3rd order current deposition for CPU and GPU

RemiLehe:
It seems that there was a small typo in the current deposition routines. This pull request fixes this.

Note that this bug only affects particles that are close to the axis (since it is inside an if condition with `ir_flip == -2`)


MKirchen:
Good catch! Sorry, I think I must have introduced that bug at some time before...

PR #137, opened 2017-11-05
Simplify current deposition on CPU and generalize it for multi-mode

RemiLehe:
The current deposition code for CPU is currently hard-coded for 2 modes, and a bit difficult to read. This pull request simplifies it and allows to use  an arbitrary number of modes. 

Importantly, those changes lead to slightly better performance ; here is a plot of the performance of the master branch (solid line) and of this PR (dashed line):
screen shot 2017-11-05 at 12 29 43 pm

Here is a summary of the changes:
- In order allow the use of arbitrary number of modes, the array that duplicate the charge/current for each thread (`rho_global`, `j*_global`) have an additional axis for the azimuthal mode. Inside the deposition kernels, loops over the modes have been introduced.
- In order to have the code work with loops over azimuthal mode, all the intermediate named variables (like `R_m1_01`) have been suppressed (although they are useful on GPU, where they accumulate the field of many particle before an atomic, on CPU they were not playing such a role and were not actually useful).
- Also, in order to simplify the code, all the `if` conditions that check whether the particle is a few cells from the boundaries (and execute corresponding specific code) have been suppressed. Instead, the array `rho_global`, `j*_global` have "deposition guard cells`, so that these `if` conditions are not needed anymore (no out-of-bound for particles that are close to the boundary). These deposition guard cells are added to the corresponding regular cell of the regular arrays `rho`, `j` during the `sum_reduce` kernel.

**Note:** When reviewing, I would recommend **not** looking at the diff of `threading_methods.py` since the changes are extensive. Instead, it is better to just have a look at the new `threading_methods.py` alone (also since it is now much shorter)


MKirchen:
Nice PR. I like the reorganization pf the code. Especially nice is the strategy to use a larger grid for the guard regions and fold them afterwards. I still wonder if we should try a similar approach (although quite different) for the GPU deposition.

One question for the CPU case: As far as I know, numba now supports to have a parallel np.sum with axis argument support. At some point (not in this PR) maybe we can replace the custom reduce with this numba inplementation. Or is this not possible with the guard cell folding approach?

RemiLehe:
Thanks for reviewing this!
Reguarding support of `np.sum`: Yeah, I think we would need to split the kernel in 2, i.e. do the folding of the guard cells separately from the sum.

RemiLehe:
I opened an issue for this: #138

PR #141, opened 2017-11-11
Make the exchange of guard cells (for rho and J) more robust

RemiLehe:
Right now some of the code relies explicitly on the fact that rho is never exchanged. While this works, it is not very robust (e.g. if at some point, or in some configuration, we decide to modify the exchange pattern).

Therefore, I introduced a set of flags that record whether the fields have been exchanged. I also modified the function `deposit` so that there is an option to automatically exchange the fields (even for `rho`, although it is not used right now): I think it makes things a bit clearer in the main PIC loop.

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -166,6 +166,11 @@ class Fields(object) :
                                 use_galilean=self.use_galilean,
                                 use_cuda=self.use_cuda ) )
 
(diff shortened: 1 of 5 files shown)


RemiLehe:
Note: if this gets merged into `dev`, then I'll propagate the changes into the branch `cross-deposition`.

PR #143, opened 2017-11-14
Allow FBPIC to run without MPI

RemiLehe:
In many use cases, FBPIC does not really require MPI (MPI is only required for parallel simulations or parallel parameter scans). Therefore, it might be good to allow FBPIC to run without mpi4y (in case some users experience issues with installing mpi4py.)

This pull request removes `mpi4py` from the requirements, and defines a dummy replacement for the places in the code where an MPI communicator is explicitly used. By default, the documentation still recommends that mpi4py be installed, since most users will not experience issues with this.

Feel free to comment/suggest alternatives, as I am not sure that this is the most elegant solution...

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -7,7 +7,7 @@ It defines the structure necessary to implement the boundary exchanges.
 """
 import numpy as np
 from scipy.constants import c
-from mpi4py import MPI as mpi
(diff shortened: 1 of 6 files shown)


RemiLehe:
Note: I ran the automated tests on my laptop after **uninstalling** `mpi4py` and all the tests that do not require mpi4py run fine. Also, I ran the code on GPU without an error.

PR #149, opened 2017-11-21
[WIP] Multimode gathering

RemiLehe:
This PR implements the gather operation for an arbitrary number of modes.
It uses the fact that numba can compile with an arbitrary number of arrays, if these arrays are stored in a tuple. 

However, this feature does not work with `parallel=True` with numba 0.36. Therefore I had to enforce again the restriction on numba 0.36.


RemiLehe:
It seems that the performance is bad when using the tuples approach with 2 modes: in the plot below, the solid line is the current `dev` branch and the dashed line is this PR.
screen shot 2017-12-15 at 8 22 19 am

RemiLehe:
Because the performance is bad with tuples, and because it will not work with numba 0.36 (there is a bug for tuples and parallel=True), I closed this branch and will reimplement it by duplicating the kernels.

PR #150, opened 2017-11-24
Generalize laser injection

RemiLehe:
This pull request:

- separates the API for the definition of the laser profile and the API for the injection of the laser (through an antenna or through direct deposition). The profiles are classes that should have a method that returns the transverse Ex and Ey fields (no more than this is needed: the code calculates the other fields automatically ; see below). For the moment only the Gaussian profile is defined, but more profiles will be defined in the future (through other pull requests).

- generalizes the direct deposition so that it works for arbitrary laser profiles

More precisely, the previous implementation had several issues:

- The decomposition of the laser profile into azimuthal modes was explicitly calculated by hand for the Gaussian pulse. This does not generalize well to other laser profiles (one would have to calculate the azimuthal decomposition by hand for each profile). In the new implementation, the azimuthal decomposition is performed numerically, a so would work for any profile.

- Similarly, the Ez field was calculated by hand for a particular laser profile, in the previous implementation. In the current implementation, the Ez field is obtained entirely numerically, from div(E) = 0. Note that the new implementation also works with MPI decomposition. (The fields are gathered into a global grid in order to perform this global operation, using a very similar method to that used for bunch initialization ; see PR #127 "Correct space-charge calculations for multi-proc simulation")

- Finally, in the previous implementation, the B field have to be explicitly calculated by hand. This is now automatically calculated numerically from the Maxwell equations (assuming the pulse is initialized in vacuum)


RemiLehe:
Also: I realized that several example scripts and laser tests used ~10 points per wavelengths. This definitely not good practice and dates from the time that the code was slow on CPU. This PR also increases the number of points per wavelength to 20.

fbpic/lpa_utils/laser/laser.py
```diff
@@ -85,84 +145,11 @@ def add_laser( sim, a0, w0, ctau, z0, zf=None, lambda0=0.8e-6,
...
-
-    # Add the Ez fields to the mode m=1 (linearly polarized laser)
-    fld.interp[1].Ez +=  0.5  * E0 * exptheta * profile_Ez
-    fld.interp[1].Bz +=  0.5j * prop * E0/c * exptheta * profile_Ez
-
-    # Up to now, only the interpolation grid was modified.
-    # Now convert the fields to spectral space.
-    if update_spectral:
-        fld.interp2spect('E')
-        fld.interp2spect('B')
```
RemiLehe: The function `add_laser` is kept for backward-compatibility, but is now much shorter. 

Also, it illustrates how one would initialize a laser with the new API. This API could still change before the next release.

soerenjalas: Should we maybe already change the example scripts to this API or do you want to wait until we decided on a final API?

RemiLehe: Yeah, I would rather wait until we have a definitive API. This PR just introduces the necessary mechanism for separating the laser profile from the injection, but not much thought was given to the API.

PR #151, opened 2017-11-25
Added version number in print message

RemiLehe:
I think it is useful (both for users and developers) to know which version of FBPIC is being run. One suggestion would be to have the version printed at the beginning of the simulation, like so:
```
*** Cuda not available for the simulation.
*** Performing the simulation on CPU.

Running FBPIC 0.6.2 on CPU with 1 proc.
```

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -643,9 +643,9 @@ def print_simulation_setup( comm, use_cuda, use_threading ):
     """
     if comm.rank == 0:
         if use_cuda:
-            message = "\nRunning FBPIC %s on GPU " %__version__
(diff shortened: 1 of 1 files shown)


RemiLehe:
Or maybe it is better to print 
```
Running fbpic-0.6.2 on CPU with 1 proc.
```
to match the name the font of the python package (`fbpic`, not `FBPIC`) and to make it clearer (with the dash) that the number indicates the version. @MKirchen @soerenjalas : any opinion? I can modify the PR accordingly.

MKirchen:
I like both options. You can decide.

PR #158, opened 2017-12-04
Refactored printing of simulation setup and the progress bar

### LINKED ISSUE #66: Add a general print function that displays simulation information
At the beginning of the loop, the code should print key information about the simulation, including:
- The number of guard cells
- The exchange period
- The resolution and timestep.

Feel free to suggest other important information. However, bear in mind that this should remain concise ; only useful information should be printed.

### LINKED ISSUE #94: Time estimator of the PIC loop: do not use first step
The estimator that computes the remaining time uses the time taken in the first step for the calculation. However, the first step is special since it includes JIT compilation. It should not be included.

### LINKED ISSUE #156: Improve printing of MPI processes
Right now, the following code prints information about the number of MPI processes and the number of threads used:
```
    if comm.rank == 0:
        if use_cuda:
            message = "\nRunning fbpic-%s on GPU " %__version__
        else:
            message = "\nRunning fbpic-%s on CPU " %__version__
        message += "with %d proc" %comm.size
        if use_threading and not use_cuda:
            message += " (%d threads per proc)" %numba.config.NUMBA_NUM_THREADS
        message += ".\n"

        print( message )
```

We should change it to the following for better readability:
(Note that I exchanged "proc" with "MPI processes")

```
    if comm.rank == 0:
        if use_cuda:
            message = "\nRunning fbpic-%s on GPU " %__version__
        else:
            message = "\nRunning fbpic-%s on CPU " %__version__
        message += "with %d MPI processes" %comm.size
        if use_threading and not use_cuda:
            message += " (%d threads per MPI process)" %numba.config.NUMBA_NUM_THREADS
        message += ".\n"

        print( message )
```

MKirchen:
General changes:
- The printing functions of main.py have been moved to a new file print_utils.py (makes main.py a little bit cleaner)

Enhancements in the printing at the beginning of the simulation:
- The simulation object takes a new argument "simulation_info" that specifies the level of detail for printing simulation information (0 = no information, 1 = basic information (default), 2 = detailed information, such as n_order/n_guard/exchange_period...)
- Prints version of FBPIC after the initialization of the simulation object
- Slight refactoring of the info about the computing architecture used: CPU / GPU / MPI / threads per process

Changes to the progress bar:
- During the first iteration a warning about the JIT compilation overhead is shown instead of the progress bar
- The first iteration is not considered by the calculation of the estimated time anymore
- A moving window-like calculation is used to get the average time per step (no averaging over all steps anymore) - the progress bar does not show an ETA until all steps of the window size are done
- The time taken by the last step is printed (This is useful to estimate current performance)
- Slightly new default style
- The summary of the total runtime at the end of the step function has been moved to a separate function in print_utils.py

Closes #156 
Closes #66 
Closes #94

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -10,7 +10,6 @@ This file steers and controls the simulation.
 # (This needs to be done before the other imports,
 # as it sets the cuda context)
 from fbpic.mpi_utils import MPI
-import numba
 # Check if threading is available
 from .threading_utils import threading_enabled
 # Check if CUDA is available, then import CUDA functions
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -20,10 +20,9 @@ if cuda_installed:
     mpi_select_gpus( MPI )
 
 # Import the rest of the requirements
-import time
+import numba
 from scipy.constants import m_e, m_p, e, c
-from .print_utils import print_simulation_setup, \
-            print_runtime_summary, progression_bar
+from .print_utils import ProgressBar, print_simulation_setup
 from .particles import Particles
 from .lpa_utils.boosted_frame import BoostConverter
 from .fields import Fields
@@ -49,7 +48,7 @@ class Simulation(object):
                  use_cuda=False, n_guard=None, n_damp=30, exchange_period=None,
                  boundaries='periodic', gamma_boost=None,
                  use_all_mpi_ranks=True, particle_shape='linear',
-                 simulation_info=1 ):
+                 verbose_level=1 ):
         """
         Initializes a simulation, by creating the following structures:
 
@@ -190,21 +189,27 @@ class Simulation(object):
             Possible values are 'cubic', 'linear'. ('cubic' corresponds to
             third order shapes and 'linear' to first order shapes).
 
-        simulation_info: int, optional
-            Print information about the simulation setup.
+        verbose_level: int, optional
+            Print information about the simulation setup after
+            initialization of the Simulation class.
             0 - Print no information
             1 (Default) - Print basic information
             2 - Print detailed information
         """
         # Check whether to use CUDA
         self.use_cuda = use_cuda
-        if (use_cuda==True) and (cuda_installed==False):
+        self.cuda_installed = cuda_installed
+        if (self.use_cuda==True) and (self.cuda_installed==False):
             # Print warning if use_cuda = True but CUDA is not available
             print('*** Cuda not available for the simulation.')
             print('*** Performing the simulation on CPU.')
             self.use_cuda = False
         # CPU multi-threading
         self.use_threading = threading_enabled
+        if self.use_threading:
+            self.cpu_threads = numba.config.NUMBA_NUM_THREADS
+        else:
+            self.cpu_threads = 1
 
         # Register the comoving parameters
         self.v_comoving = v_comoving
@@ -213,14 +218,15 @@ class Simulation(object):
             self.use_galilean = False
 
         # When running the simulation in a boosted frame, convert the arguments
-        self.gamma_boost = gamma_boost
         uz_m = 0.   # Mean normalized momentum of the particles
-        if self.gamma_boost is not None:
-            boost = BoostConverter( self.gamma_boost )
-            zmin, zmax, dt = boost.copropag_length([ zmin, zmax, dt ])
-            p_zmin, p_zmax = boost.static_length([ p_zmin, p_zmax ])
-            n_e, = boost.static_density([ n_e ])
-            uz_m, = boost.longitudinal_momentum([ uz_m ])
+        if gamma_boost is not None:
+            self.boost = BoostConverter( gamma_boost )
+            zmin, zmax, dt = self.boost.copropag_length([ zmin, zmax, dt ])
+            p_zmin, p_zmax = self.boost.static_length([ p_zmin, p_zmax ])
+            n_e, = self.boost.static_density([ n_e ])
+            uz_m, = self.boost.longitudinal_momentum([ uz_m ])
+        else:
+            self.boost = None
 
         # Initialize the boundary communicator
         self.comm = BoundaryCommunicator( Nz, zmin, zmax, Nr, rmax, Nm, dt,
@@ -283,7 +289,7 @@ class Simulation(object):
         self.laser_antennas = []
 
         # Print simulation setup
-        print_simulation_setup( self, level=simulation_info )
+        print_simulation_setup( self, verbose_level=verbose_level )
 
     def step(self, N=1, correct_currents=True,
              correct_divE=False, use_true_rho=False,
@@ -328,9 +334,7 @@ class Simulation(object):
 
         # Initialize variables to measure the time taken by the simulation
         if show_progress and self.comm.rank==0:
-            start_time = time.time()
-            prev_time = start_time
-            avg_time_per_step = 0.
+            progress_bar = ProgressBar( N )
 
         # Send simulation data to GPU (if CUDA is used)
         if self.use_cuda:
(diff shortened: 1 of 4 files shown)


fbpic/print_utils.py
```diff
@@ -0,0 +1,234 @@
...
+                    %sim.comm.gamma_boost
+                if sim.use_galilean:
+                    message += '\nGalilean frame: Yes'
+                else:
+                    message += '\nGalilean frame: No'
+            else:
+                message += '\nBoosted frame: False'
+        message += '\n'
+        print( message )
+    if level == 2:
```
RemiLehe: I think I would put this as level 1 ; but I do not have a very strong feeling on this: up to you.

fbpic/print_utils.py
```diff
@@ -0,0 +1,234 @@
...
+    # Print progress bar
+    if i == 0:
+        # Let the user know that the first step is much longer
+        sys.stdout.write('\r' + \
+            'Just-In-Time compilation (up to one minute) ...')
+        sys.stdout.flush()
+    else:
+        # Print the progression bar
+        nbars = int( (i+1)*1./Ntot*Nbars )
+        sys.stdout.write('\r' + nbars*char )
```
RemiLehe: I have a slight personal preference for having the square brackets around the nbars characters (esp. when the character is the simple '-'), like this:
```
[--------             ]
```
Any strong opinion on this?

fbpic/main.py
```diff
@@ -312,13 +326,11 @@ def step(self, N=1, correct_currents=True,
...
             raise ValueError('correct_divE cannot be used in multi-proc mode.')
 
-        # Let the user know that the first step is much longer
-        if show_progress and (self.comm.rank==0):
-            print('Performing %d PIC iterations:' %N )
-            print(' - Just-In-Time compilation (up to one minute) ...')
-
-        # Measure the time taken by the PIC cycle
-        measured_start = time.time()
+        # Initialize variables to measure the time taken by the simulation
```
RemiLehe: I like how the modifications below clean the code in the main loop. 
One possibility to make it even cleaner would be to write a class (e.g. `TimePrinter`, in `printing_utils.py`) and create an instance of that class, here, that records the `start_time`, `prev_time` and `avg_time_per_step`, so that these variables do not appear explicitly in the code for the `step` function.
Then this instance could have the method `progression_bar` and `print_runtime_summary`.

fbpic/print_utils.py
```diff
@@ -0,0 +1,234 @@
...
+    duration: float, seconds
+        The total time taken by the step loop
+    """
+    avg_tps = (duration / N)*1.e3
+    m, s = divmod(duration, 60)
+    h, m = divmod(m, 60)
+    print('\nTime taken (with compilation): %d:%02d:%02d' %(h, m, s))
+    print('Average time per iteration ' \
+          '(with compilation): %d ms\n' %(avg_tps))
+
```
RemiLehe: I wonder if we should remove the `print_gpu_meminfo` functions. It seems that in most cases `nvidia-smi` would do the job, no?

fbpic/main.py
```diff
@@ -312,13 +326,11 @@ def step(self, N=1, correct_currents=True,
         if self.comm.size > 1 and correct_divE:
             raise ValueError('correct_divE cannot be used in multi-proc mode.')
 
-        # Let the user know that the first step is much longer
-        if show_progress and (self.comm.rank==0):
-            print('Performing %d PIC iterations:' %N )
```
RemiLehe: I would have a slight preference for keeping `print('Performing %d PIC iterations:' %N )` (this would then be moved to the `printing_utils.py`).
What do you think?

RemiLehe:
Thanks a lot for this pull request!
I really like how it cleans the code in `main.py`! And the idea of the different verbosity levels is great ; `level=2` gives a really nice summary of the relevant information!

fbpic/main.py
```diff
@@ -48,7 +48,8 @@ def __init__(self, Nz, zmax, Nr, rmax, Nm, dt, p_zmin, p_zmax,
                  v_comoving=None, use_galilean=True, initialize_ions=False,
                  use_cuda=False, n_guard=None, n_damp=30, exchange_period=None,
                  boundaries='periodic', gamma_boost=None,
-                 use_all_mpi_ranks=True, particle_shape='linear' ):
+                 use_all_mpi_ranks=True, particle_shape='linear',
+                 simulation_info=1 ):
```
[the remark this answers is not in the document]
MKirchen: done

fbpic/main.py
```diff
@@ -188,10 +189,17 @@ def dens_func( z, r ) ...
             Set the particle shape for the charge/current deposition.
             Possible values are 'cubic', 'linear'. ('cubic' corresponds to
             third order shapes and 'linear' to first order shapes).
+
+        simulation_info: int, optional
```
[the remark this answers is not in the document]
MKirchen: done

fbpic/main.py
MKirchen: I would still remove this line. It would only be printed if show_progress == True and in this case the total number of iterations are printed by the progress bar anyways. So this would be redundant.

fbpic/main.py
MKirchen: Great idea, I created a class ProgressBar which handles all of this

fbpic/print_utils.py
```diff
@@ -0,0 +1,234 @@
...
+    level: int, optional
+        Level of detail of the simulation information
+        0 - Print no information
+        1 (Default) - Print basic information
+        2 - Print detailed information
+    """
+    if level > 0 and sim.comm.rank == 0:
+        # Print version of FBPIC
+        message = '\nFBPIC (fbpic-%s)\n'%__version__
```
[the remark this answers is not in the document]
MKirchen: done

fbpic/print_utils.py
MKirchen: I would keep it at level 2. for level 1, it should be sufficient for the user to know that the simulation is done on GPU with N number of MPI processes. (This is already printed in the case of level 1). Also this then gives consistent printing behaviour for either CPU or GPU simulations (in case of level = 1)

fbpic/print_utils.py
MKirchen: The new default style of the progress bar does not look nice together with the brackets. (I hope you like the new default style). Therefore, I would remove the brackets.

Another option would be to use " | " as an indicator for the start and end of the progress bar. This looks okay with the new default style.

fbpic/print_utils.py
MKirchen: You are right, but I would still keep this function for the moment. It might be useful at some point for debugging as we can use it to quickly track the memory consumption at different places in the loop during execution.

MKirchen:
Note: In the last commit, I added some additional registered variables to the simulation object, which I think should be registered there as they contain important information about the simulation. This also helped to remove all the not-so-nice imports in print_utils.py

fbpic/main.py
RemiLehe: OK, sounds good.

RemiLehe:
OK, thanks a lot for implementing the changes! This really looks nice with the new ProgressionBar class!

I have a one more comment (sorry for being picky!), then it's good to good:

I agree with most of the new attributes that you are registering in the class (e.g. `boost`, `boundaries`) because they are really specific to the simulation. However, `MPI`, `mpi_installed` and `cuda_installed` are global variables that are not specific to the simulation (in other words, even if we initialize two simulations in the same script, these parameters will always be the same for both simulations). For this reason, it is a bit counter-intuitive to me to register them in the class, as it creates duplicate information (and potentially confusions). If it's okay with you, I would prefer the old import system for these parameters.

MKirchen:
Ready to merge

RemiLehe:
Nice! Thanks for this PR!

PR #161, opened 2017-12-06
Reduce the number of Hankel transforms per iteration

RemiLehe:
In the current dev branch, we perform 16 Hankel transform and 16 Fourier transform per azimuthal mode and per iteration. Within those 16 transforms, 12 transforms are done because we do a `spect2interp` and `interp2spect` of the fields E and B (2 * 2 fields * 3 components).

The reason for performing these back-and-forth transformations is that we have to perform some operations on the fields in spectral space (Maxwell push) and some operations in real space (MPI exchanges and damping of the open boundaries). 

However, because the MPI exchanges and damping of the open boundaries is purely along z, updating the spectral fields can be done only by a succession an inverse and forward Fourier transform (no Hankel transform involved). Then updating the fields in interpolation space requires an additional Hankel + Fourier transform.

This effectively replaces the 16 Hankel and 16 Fourier transforms by 10 Hankel and 22 Fourier transforms (per mode, per iteration). Because the Fourier transforms are almost always much faster than the Hankel transforms, this results in a speedup of the simulation.

*Note: I am definitely not the first to use this trick! @hightower has been using it for a while in his code [chimera](https://github.com/hightower8083/chimera).*

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -705,10 +705,10 @@ class BoundaryCommunicator(object):
         add_buffers_to_particles( species, float_recv_left, float_recv_right,
                                     uint_recv_left, uint_recv_right )
 
-    def damp_guard_EB( self, interp ):
+    def damp_EB_open_boundary( self, interp ):
         """
-        Apply the damping shape in the right and left guard cells.
-        Damp the fields E and B in the guard cells.
+        Damp the fields E and B in the damp cells, at the right and left
+        of the *global* simulation box.
 
         Parameter:
         -----------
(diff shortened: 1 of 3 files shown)


MKirchen:
Nice, looking forward to this PR!

fbpic/boundaries/boundary_communicator.py
```diff
@@ -705,10 +705,10 @@ def exchange_particles_aperiodic_subdomain(self, species, fld, time ):
         add_buffers_to_particles( species, float_recv_left, float_recv_right,
                                     uint_recv_left, uint_recv_right )
 
-    def damp_guard_EB( self, interp ):
```
RemiLehe: Note: I renamed this function to be more explicit about what it does (esp. because the guard cells inbetween two MPI processes are not damped by this function)

RemiLehe:
I performed the usual 2000 x 400 benchmark (timing the complete PIC, not just the transforms) on a GTX 1080 Ti GPU:
- Current dev: 325 ms/iteration
- This PR: 275 ms/iteration

MKirchen:
Nice!

hightower8083:
wow, 18 percent !

PR #162, opened 2017-12-07
Fix checkpoint restart

MKirchen:
The checkpoint / restart feature was broken:

- In load_species the weights were multiplied by the charge q. This should not be done anymore.
- If a diag_period was at the same time as a checkpoint_period, a wrong diagnostic was written at the first step (i_step == 0) after the restart (wrong charge/currents). With this PR, a check is performed based on a new attribute (is_restarted) of the Simulation object.

Furthermore, the user is informed about the restart and at which iteration the simulation is restarted.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -259,6 +259,8 @@ class Simulation(object):
         # Register the time and the iteration
         self.time = 0.
         self.iteration = 0
+        # Register variable that keeps track of restarting
+        self.is_restarted = False
         # Register the filtering flag
         self.filter_currents = filter_currents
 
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for spotting this issue! And thanks for fixing the particle weight.

Regarding the incorrectness of the first diagnostic: I am a bit reluctant to fix it in this way, essentially for 2 reasons:
- The first diagnostic of a regular simulation (without a restart) will also have a wrong charge/current (because the charge and currents have not been deposited yet)
- This changes the requests of the user (e.g. one diagnostic every 100 timestep) without warning him ; therefore he might think that there is a bug.
However, I do not have a much better solution to offer. 

For now, since the problem with the particle weight is quite serious and easy to fix: would you be okay with doing a PR (or modify this PR) to only fix this problem?

MKirchen:
I agree that it might not be the perfect solution, but I dont think that it will confuse the user. The total number of diagnostics will be consistent with this PR and will look the same way as when doing the simulation without checkpoints.

For example: If the user sets the checkpoint period to 50 and the diagnostic period to 10 and performs 200 steps. Then there will be a diagnostic and a checkpoint at step 100. 
If the user now restarts the simulation:
- Current dev: The diagnostic 200 will be overwritten with a diagnostic that does not contain any charge or currents
- This PR: The code recognizes that it was restartet and it will not overwrite the diagnostic 200

Or do I miss something here?

RemiLehe:
As discussed off-line, I will do a new implementation of this PR.

PR #165, opened 2017-12-08
Close #164 New printing functions break when mpi is not installed, for `verbose_level=2`

### LINKED ISSUE #164: New printing functions break when mpi is not installed, for `verbose_level=2`
The code breaks when printing the gpu information: 
```
 File "fbpic_script.py", line 92, in <module>
    use_all_mpi_ranks=False, initialize_ions=False, verbose_level=2 )
  File "/global/home/users/rlehe/miniconda3/lib/python3.6/site-packages/fbpic-0.6.2-py3.6.egg/fbpic/main.py", line 291, in __init__
    print_simulation_setup( self, verbose_level=verbose_level )
  File "/global/home/users/rlehe/miniconda3/lib/python3.6/site-packages/fbpic-0.6.2-py3.6.egg/fbpic/print_utils.py", line 224, in print_simulation_setup
    sim.comm.mpi_comm.barrier()
AttributeError: 'NoneType' object has no attribute 'barrier'
```

RemiLehe:
This uses the fact that we have usually 2 distinct MPI "communicators" in an FBPIC run:
- The global `MPI.COMM_WORLD`: which is either the global `mpi4py` communicator, or an instance of `DummyCommunicator` (when `mpi4py` is not installed).
- The communicator that the simulation actually uses: `sim.comm.mpi_comm`: which is either the global `mpi4py` communicator, or `None` (for single-proc simulations ; e.g. when `use_all_mpi_ranks` is False) ; at some point, this could also be an MPI group.

When printing the GPU names, it would not be ideal to call Barrier with the **global** `MPI.COMM_WORLD`. This is because the code for printing the GPU names is called when creating the `Simulation` object, and because when running a set of simulations with `use_all_ranks=False`, different MPI ranks might, in principle, not create the same number of `Simulation` objects. (although it is a very hypothetical scenario.)

Instead, in this PR, we use an `MPI gather` on the local communicator `sim.comm.mpi_comm`, to gather the strings containing the GPU name from each proc in the local communicator (and in the end only the local proc 0 prints the information). This ensures that, when using `use_all_ranks=False`, the different parametric simulations will not interfere with each other (although they might still print their message to the screen in an unordered fashion ; but at least we avoid any potential deadlock).

Another advantage of using `gather`: with some version mpi4py, the prints from the other procs than proc 0 are only appended to the terminal *at the end of the simulation* (i.e. the GPU information will appear after the progression bar in this case). Having proc 0 print the gathered information fixes this issue.

### DIFF
--- a/fbpic/print_utils.py
+++ b/fbpic/print_utils.py
@@ -154,7 +154,7 @@ def print_simulation_setup( sim, verbose_level=1 ):
         1 (Default) - Print basic information
         2 - Print detailed information
     """
-    if verbose_level > 0 and sim.comm.rank == 0:
+    if verbose_level > 0:
         # Print version of FBPIC
         message = '\nFBPIC (%s)\n'%__version__
         # Basic information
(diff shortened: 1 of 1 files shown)


RemiLehe:
Also, the GPU information from each MPI rank now appears directly after printing whether CUDA is installed.

MKirchen:
Thanks!

Issue #163, opened 2017-12-07
Remove numba=0.34 restriction


MKirchen:
It seems like the threading works fine with numba 0.36 (tested with periodic plasma wave test). We should remove the restriction that checks if numba 0.34 is installed and instead require numba 0.36 for the installation.

RemiLehe:
Agreed! I also tested with numba 0.36 and had no issue either.

RemiLehe:
Note: we should still have a check in `threading_utils.py` that tests whether the numba version is either 0.34 or 0.36 (since 0.35 fails).

MKirchen:
Ill prepare a small PR

PR #167, opened 2017-12-08
Allow CPU multi-threading with numba 0.36

MKirchen:
The newest version of numba (0.36) works fine with FBPIC multi-threading.
The restriction in threading_utils.py has been modified such that numba 0.34 and 0.36 can be used together with multi-threading. 
When numba 0.34 or 0.36 are not installed, the user is recommended to install numba 0.34 (as 0.36 is not yet included in the official anaconda channel)

Note: We should remove this restriction completely in the future and just require a minimum numba version.

### DIFF
--- a/fbpic/threading_utils.py
+++ b/fbpic/threading_utils.py
@@ -23,13 +23,15 @@ if threading_enabled:
     try:
         # Try to import the threading function prange
         from numba import prange as numba_prange
-        # (Temporarily) check that numba is version 0.34 (other versions fail)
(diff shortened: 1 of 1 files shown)


MKirchen:
Closes #163

RemiLehe:
Thanks! I agree with only requiring a minimal version in the future (0.36, I guess). This would allow use to clean the code and replace the while loops in threaded kernels with for loops (which numba 0.34 did not correctly support)

PR #175, opened 2017-12-15
Replace the function `merge_buffers_to_particles` by a faster set of function

RemiLehe:
For some reason, the function `merge_buffers_to_particles` is very slow. For instance, in the usual 2000x400 benchmark with linear shape factors, the cost of this function can be of the same order as the gathering and deposition functions!

This is very surprising since the function `split_particles_to_buffers` is almost identical, but takes 20x less time per call!

In any case, I now replaced the function `merge_buffers_to_particles` by a set of simpler functions. The time taken by these different functions is now negligible. For instance, on the 2000x400 benchmark with a GTX 1080 Ti GPU, the `dev` branch takes 56 s while this PR takes 51 s.

### DIFF
--- a/fbpic/boundaries/particle_buffer_handling.py
+++ b/fbpic/boundaries/particle_buffer_handling.py
@@ -454,11 +454,15 @@ def add_buffers_gpu( species, float_recv_left, float_recv_right,
         These arrays are always on the CPU (since they were used for MPI)
     """
     # Get the new number of particles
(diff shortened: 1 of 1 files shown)


RemiLehe:
Note: the automated tests pass on GPU, with this PR.

MKirchen:
Cool improvement. Interesting that it was so slow before.

PR #181, opened 2018-01-05
Reduce duplication in gathering kernels by using more elementary functions

RemiLehe:
In the code for field gathering, there is a lot of code duplication:
- the gathering and weighting of E is very similar to gathering and weighting of B
- the gathering for mode 0 is very similar to the gathering for mode 1
- the code for CPU is very similar to the code for GPU

For this reason, I wrote a single function that does the gathering and weighting for an arbitrary field (E or B), arbitrary mode, and for both CPU and GPU (with one version for linear shape and one version cubic shape). This elementary function is then used in the CPU and GPU kernels (it is compiled either `cuda` or `numba` ; a similar approach was used in the ionization code.)

Also, I got rid of the `iz` and `ir` arrays for cubic deposition (as they are very easy to recalculate on the fly, just before accessing the corresponding fields.)

The tests pass on CPU and GPU. 

Here are the timings on GPU (only for the gathering kernel, time per call, measured on a GTX 1080):

|     | Linear gathering | Cubic gathering | 
|-----|------------------|-----------------|
| dev | 20.920ms         | 120.87ms        |
| PR  | 20.723ms         | ~~77.80ms~~ 79.72ms       |

I am quite surprised by the speed-up in the cubic gathering (~~I did change the cubic code a little to avoid calculating `Sz*Sr` many times (see the variable `weight`)~~). @MKirchen @soerenjalas Would you have time to do some profiling on a K80 or P100?

Here are the timing on a 12-core CPU, as a function of the number of threads (the dashed line corresponds to this PR, the solid line corresponds to the `dev` branch):
screen shot 2018-01-05 at 3 19 46 pm

So in summary: 
- For cubic shape, the new code is much faster on GPU, and just lightly slower for CPU.
- For linear shape, the new code has about the same speed on GPU, but is significantly slower on CPU. However, this is mitigated for a large number of threads. Moreover, for linear shape, the gathering is usually not the most costly part of the PIC loop.

For this reason (and for all the simplifications that this PR makes in the code), my opinion would be that this is worth merging. @MKirchen @soerenjalas : Any opinion?

### DIFF
--- a/fbpic/particles/gathering/cuda_methods.py
+++ b/fbpic/particles/gathering/cuda_methods.py
@@ -9,9 +9,12 @@ on the GPU using CUDA.
 from numba import cuda, float64, int64
 import math
 # Import inline functions
-from .inline_functions import add_linear_gather_for_mode
+from .inline_functions import \
+    add_linear_gather_for_mode, add_cubic_gather_for_mode
 # Compile the inline functions for GPU
 add_linear_gather_for_mode = cuda.jit( add_linear_gather_for_mode,
+                                        device=True, inline=True )
+add_cubic_gather_for_mode = cuda.jit( add_cubic_gather_for_mode,
                                         device=True, inline=True )
 
 # -----------------------
@@ -167,7 +170,7 @@ def gather_field_gpu_linear(x, y, z,
         Ez[i] = Fz
 
         # B-Field
-        # ----------------------------
+        # -------
         # Clear the placeholders for the
         # gathered field for each coordinate
         Fr = 0.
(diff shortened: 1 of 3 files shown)


MKirchen:
Wow, I always thought about the possibility to use more registers (bad) to precalculate some more variables (good), but I thought that this would not make a difference compiler-wise. Its amazing how much speed-up we can get fromt this PR. Nice!

RemiLehe:
Yes, but unfortunately I am realizing that this code seems to be significantly slower on CPU (no idea why... Maybe function call overhead...)
I am working on it (hence the [WIP] flag).

RemiLehe:
I updated the code for cubic shapes in the PR (see updated description at the top).
This is ready for review.

PR #182, opened 2018-01-07
Close #139 Simplify/optimize the code for current deposition on GPU

### LINKED ISSUE #139: Simplify/optimize the code for current deposition on GPU
It would be interesting to take a similar approach to PR #137

RemiLehe:
This pull request is a proposition to slightly change the way in which we sort the particles on the GPU, in order to simplify the deposition kernels.

More precisely, in the current `dev` branch, the particles are sorted according to the index of their lower grid point in z and r as shown here with tuples of `(iz_lower, ir_lower)`. (In the code, these tuples are actually translated into a single index `ir_lower + Nr*iz_lower`, but this is a detail.)
screen shot 2018-01-09 at 10 27 35 pm
Notice that the particles that are below the first radial grid cell do not have a lower grid point in r, and so, **currently**, they are put in the same "bag" as the particles that are inbetween the first radial grid point and second radial grid point. (The same problem does not occur in `z` because of the periodicity in z.)

Then, the deposition kernel on GPU go through the particles in each colored area and deposits charge/current to the surrounding grid points. However, because the particles below and above the first radial grid cell do not deposit charge to the same grid points and in the same way, the deposition kernel need a lot of if conditions to treat this special case.

Instead, I suggest that we slightly change the index according to which the particles are sorted and use, instead, the upper index:
screen shot 2018-01-09 at 10 37 03 pm
The particles that are above the last grid point do not have an upper grid point, but in that case we can simply use `ir_upper = Nr`. Thus the index `ir_upper` can have values between `0` and `Nr` (included) and so the linearized index should be calculated according to `ir_upper + (Nr+1)*iz_upper`. (The same could have been done with `ir_lower` having values between `-1` and `Nr-1` but is less intuitive because of potentially negative values.)

In this case, the deposition kernels can be significantly simplified because we now know, from the sorting index (a.k.a prefix sum index) to which cells the macroparticles will deposit, without ambiguity. This is what this pull request does.

The automated tests pass on GPU.

The changes show essentially no speed-up (see table below), but allow, on the whole, to remove ~400 lines of code. In addition, I think the kernels are also conceptually slightly easier to understand.

|     | Cubic J  | Cubic rho | Linear J | Linear rho |
|-----|----------|-----------|----------|------------|
| dev | 89.438ms | 22.496ms  | 16.943ms | 5.7967ms   |
| PR  | 89.117ms | 22.413ms  | 16.832ms | 5.7674ms   |

This PR has a lot of changes in the file `cuda_methods.py`. I suggest looking at the kernel for linear deposition of rho to get the idea of the PR (since it is the simplest kernel and the other kernels are just repetitions of the same idea.)

### DIFF
--- a/fbpic/particles/deposition/cuda_methods.py
+++ b/fbpic/particles/deposition/cuda_methods.py
@@ -129,9 +129,10 @@ def deposit_rho_gpu_linear(x, y, z, w, q,
     i = cuda.grid(1)
     # Deposit the field per cell in parallel (for threads < number of cells)
     if i < prefix_sum.shape[0]:
-        # Calculate the cell index in 2D from the 1D threadIdx
-        iz_cell = int(i / Nr)
-        ir_cell = int(i - iz_cell * Nr)
+        # Calculate the lowest cell index in 2D where the particle
+        # deposits charge, from the 1D threadIdx
+        iz_lowest = int( i / (Nr+1) )
+        ir_lowest = int( i - iz_lowest * (Nr+1) ) - 1
         # Calculate the inclusive offset for the current cell
         # It represents the number of particles contained in all other cells
         # with an index smaller than i + the total number of particles in the
@@ -149,7 +150,6 @@ def deposit_rho_gpu_linear(x, y, z, w, q,
         R_m0_01 = 0.
         R_m0_10 = 0.
         R_m0_11 = 0.
-
         R_m1_00 = 0. + 0.j
         R_m1_01 = 0. + 0.j
         R_m1_10 = 0. + 0.j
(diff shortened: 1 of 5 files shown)


fbpic/particles/particles.py
```diff
@@ -617,11 +617,8 @@ def deposit( self, fld, fieldtype ) :
         # GPU (CUDA) version
         if self.use_cuda:
             # Get the threads per block and the blocks per grid
-            dim_grid_2d_flat, dim_block_2d_flat = cuda_tpb_bpg_1d(
-                                                    grid[0].Nz*grid[0].Nr,
-                                                    TPB=64 )
```
RemiLehe: It turns out that `dim_grid_2d` (defined below), was not used in the code.

fbpic/particles/deposition/cuda_methods.py
```diff
@@ -195,51 +195,43 @@ def deposit_rho_gpu_linear(x, y, z, w, q,
...
-                R_m0_00 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 0) * R_m0_scal
-                R_m0_01 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 1) * R_m0_scal
-                R_m1_00 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 0) * R_m1_scal
-                R_m1_01 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 1) * R_m1_scal
-            else:
-                R_m0_10 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 0) * R_m0_scal
-                R_m0_11 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 1) * R_m0_scal
-                R_m1_10 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 0) * R_m1_scal
-                R_m1_11 += r_shape_linear(r_cell, 1)*z_shape_linear(z_cell, 1) * R_m1_scal
-
```
RemiLehe: I replaced `shift_r` and `shit_z` below by corresponding indices (`ir0`, `ir1`, `iz0`, `iz1`). This allows to write the set of atomic adds (further below) in a more systematic way.

fbpic/particles/deposition/cuda_methods.py
```diff
@@ -195,51 +195,43 @@ def deposit_rho_gpu_linear(x, y, z, w, q,
...
             R_m1_scal = wj * exptheta_m1
 
-            # Boundary Region Shifts
-            ir_lower = int64(math.floor(r_cell))
-
             R_m0_00 += r_shape_linear(r_cell, 0)*z_shape_linear(z_cell, 0) * R_m0_scal
             R_m0_01 += r_shape_linear(r_cell, 0)*z_shape_linear(z_cell, 1) * R_m0_scal
             R_m1_00 += r_shape_linear(r_cell, 0)*z_shape_linear(z_cell, 0) * R_m1_scal
             R_m1_01 += r_shape_linear(r_cell, 0)*z_shape_linear(z_cell, 1) * R_m1_scal
-
```
RemiLehe: This is the if condition that gets suppressed (I kept the `else:` part and unindented it). The "folding-in" of the charge deposited below the axis is done in the atomics operations further below (see calculation of the index `ir0`.)

fbpic/particles/deposition/cuda_methods.py
```diff
@@ -195,51 +195,43 @@ def deposit_rho_gpu_linear(x, y, z, w, q,
...
-            cuda.atomic.add(rho_m1.real, (iz_cell+1 + shift_z, ir_cell), R_m1_01.real)
-            cuda.atomic.add(rho_m1.imag, (iz_cell+1 + shift_z, ir_cell), R_m1_01.imag)
-
-            cuda.atomic.add(rho_m0.real, (iz_cell, ir_cell+1 + shift_r), R_m0_10.real)
-            cuda.atomic.add(rho_m1.real, (iz_cell, ir_cell+1 + shift_r), R_m1_10.real)
-            cuda.atomic.add(rho_m1.imag, (iz_cell, ir_cell+1 + shift_r), R_m1_10.imag)
-
-            cuda.atomic.add(rho_m0.real, (iz_cell+1 + shift_z, ir_cell+1 + shift_r), R_m0_11.real)
-            cuda.atomic.add(rho_m1.real, (iz_cell+1 + shift_z, ir_cell+1 + shift_r), R_m1_11.real)
-            cuda.atomic.add(rho_m1.imag, (iz_cell+1 + shift_z, ir_cell+1 + shift_r), R_m1_11.imag)
```
RemiLehe: The atomics are written in a more systematic way, but are equivalent to the previous code.

MKirchen:
Nice! This really simplifies the code a lot!

PR #194, opened 2018-02-08
Automate and generalize the test for linear laser-wakefield

RemiLehe:
I modified the linear laser-wakefield test so that:
- It can test the wakefield generated in different azimuthal modes (including m=2)
- It checks the accuracy of the wakefield (profile and amplitude) automatically
- The test runs for each push (but only using m=0, in this case, for faster tests)

Regarding the first point (wakefield generated in different azimuthal modes), I modified the script so that it tests the following situations: (For each situation, the 3D image shows the laser (red-orange) and Ez in the wakefield (blue-yellow).)

- **Donut-shaped laser, polarized along theta**: in this case, both the laser and the wakefield are donut-shaped, and entirely contained in mode m=0. So only one azimuthal mode is used in the calculation. (This is the case which is automatically tested for each push.)
screen shot 2018-02-07 at 10 36 27 pm

- **Linearly-polarized Gaussian laser**: This is the standard LWFA case (laser in mode m=1, wakefield in mode m=0). This uses 2 modes.
screen shot 2018-02-07 at 10 32 14 pm

- **Linearly-polarized Laguerre-Gauss laser**: This illustrates the new capabilities of the higher-order modes. Here the laser and the wakefield are in mode m=0 and m=2. The simulation uses 3 modes.
3_modes

Although the Travis CI tests only tests the first situation (the cheapest one, computationally), one can test all three configurations by doing `python test_linear_wakefield.py`. I ran it on CPU and GPU and checked that it produced the correct wakefield, including in the case involving m=2.

Note: This makes the Travis tests longer by ~3-4 min, but I think that this is an important test to have, as it is the closest to simulations that we actually use in production.

### DIFF
--- a/tests/test_linear_wakefield.py
+++ b/tests/test_linear_wakefield.py
@@ -113,11 +113,12 @@ def Er( z, r, t) :
 # Comparison plots
 # ---------------------------
 
-def compare_wakefields(Ez_analytic, Er_analytic, grid):
+def compare_wakefields(Ez_analytic, Er_analytic, grids):
     """
     Draws a series of plots to compare the analytical and theoretical results
     """
     # Get extent from grid object
+    grid = grids[0] # First mode
     extent = np.array([ grid.zmin-0.5*grid.dz, grid.zmax+0.5*grid.dz,
                         -0.5*grid.dr, grid.rmax + 0.5*grid.dr ])
     # Rescale extent to microns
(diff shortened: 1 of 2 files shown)


RemiLehe:
Many thanks again to @xxiiri for developing the Visit plugin for openPMD, which allows to visualize these modes in 3D.

MKirchen:
amazing new test! I love it!

PR #195, opened 2018-02-09
Modify test_periodic_plasma so that it can be run in parallel

RemiLehe:
This modification was previously made on the `cross_deposition` branch.
Because this change is not specific to the `cross_deposition`, I implement it separately here.

Note that, because we currently do not have a charge-conserving algorithm for parallel simulations, this test fails when run in parallel (but is successful when run in serial).


RemiLehe:
Note: this also fixes the error with `Nz` in the current `cross_deposition` branch. 
So once this branch is merged into `dev`, we can merge again `dev` into `cross_deposition`.

PR #197, opened 2018-02-10
Remove old plotting functions

RemiLehe:
These functions date from when the code was still experimental.

I don't think that they are useful anymore, and they make the code longer/less readable. Would you be okay to remove them?

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -674,66 +674,6 @@ class InterpolationGrid(object) :
         self.Jz = self.Jz.copy_to_host()
         self.rho = self.rho.copy_to_host()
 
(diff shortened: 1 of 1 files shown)


fbpic/fields/fields.py
```diff
@@ -819,7 +753,7 @@ def __init__(self, kz_modified, kr, m, kz_true, dz, dr, use_cuda=False ) :
             self.d_kz = cuda.to_device( self.kz )
             self.d_kr = cuda.to_device( self.kr )
             self.d_field_shift = cuda.to_device( self.field_shift )
-            # NB: F is not needed on the GPU (on-the-fly variables)
```
RemiLehe: I think that this comment is outdated now.

PR #202, opened 2018-02-16
Cross deposition

RemiLehe:
This implements the 'cross-deposition' current deposition.

The cross-deposition is more local that the other current deposition scheme ('curl-free') and is thus more adapted for MPI decomposition.

However, a few additional tests are still needed for the cross-deposition. Therefore, the default current correction will still be the 'curl-free' deposition.

Note that the information on the current correction type is now stored at the `Fields` level, not at the `Simulation` level.


MKirchen:
🎉 🎉 🎉

PR #216, opened 2018-03-02
Correct round-off errors in GPU deposition

RemiLehe:
In [PR #182](https://github.com/fbpic/fbpic/pull/182), the calculation of the particle index (for sorting on the GPU) was switched from using the lower index (with `math.floor`) to using the upper index (with `math.ceil`).

Then when doing the charge/current deposition, this index is used in order to extract the radial and longitudinal index at which the particles should deposit their charge/current. However, when computing the shape factors, the radial and longitudinal index are **recomputed independently**. In theory, this could lead to discrepancies, unless these radial and longitudinal indices are computed **exactly in the same way** in the sorting algorithm and in the shape factors.

Right now, the indices **in the shape factors** are computed use the `math.floor` function. This will agree with the indices extracted from the sorting index (which used `math.ceil`) **under the assumption that** `math.ceil(x) = math.floor(x) + 1`. This is true for any `x`, except **when `x` is exactly an integer**. In practice, this indeed occurs for particles that are exactly at the boundary between two cells. In this case, there is a discrepancy between the two kind of indices, and the deposited current/charge is wrong (and shows differences with the one calculated on CPU).

In order to fix this, I simply now use the `ceil` function in the shape factor, so that the indices from sorting and indices from the shape factor are calculated in exactly the same way.

I also added an automated test that compares the deposited charge/current on CPU and GPU (running for only a few steps, and with only a few particles). This test fails with the current dev branch, but passes with this PR.

### DIFF
--- a/fbpic/particles/deposition/cuda_methods.py
+++ b/fbpic/particles/deposition/cuda_methods.py
@@ -6,7 +6,7 @@ This file is part of the Fourier-Bessel Particle-In-Cell code (FB-PIC)
 It defines the deposition methods for rho and J for linear and cubic
 order shapes on the GPU using CUDA.
 """
-from numba import cuda, int64
+from numba import cuda
 import math
 from scipy.constants import c
 import numpy as np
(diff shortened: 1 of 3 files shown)


RemiLehe:
I also removed the `int64`, since I am not sure why this was needed.

PR #217, opened 2018-03-04
Removed Barrier() statements and updated Wait() pattern in MPI communication

### LINKED ISSUE #211: Removing Barrier() in boundary communicator slows down code
As @RemiLehe already discovered, removing the Barrier() commands in the boundary communicator does slow down the code execution instead of speeding it up. This is highly unexpected as the code already Waits() before this line is executed and we should investigate if something else is wrong with the send/receive pattern that causes this behaviour.

MKirchen:
This PR removes the non-ideal global Barrier() statements that were used in the MPI communication to synchronize ranks for performance and fail-safety reasons. Also, the MPI Wait() pattern for the non-blocking communications (Isend and Irecv) have been updated. Now the code waits for requests for each of the non-blocking commands. This seems to be required for good performance and correct behaviour when not using global Barrier() synchronization.

Three different cases have been benchmarked (average time per step) with the script attached below.
Setup: JURECA, 2x4 = 8xK80 on 2 nodes with 8 MPI processes.

1) Barrier() and old Wait() pattern (master):    94 ms 
2) no Barrier() and old Wait() pattern:    168 ms and MPI error at the end
3) no Barrier() and new Wait() pattern (this PR):    90 ms

[fbpic_mpi_benchmark.py.zip](https://github.com/fbpic/fbpic/files/1778785/fbpic_mpi_benchmark.py.zip)

Closes #211

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -510,9 +510,6 @@ class BoundaryCommunicator(object):
                 self.exchange_domains(
                     vec_send_left, vec_send_right,
                     vec_recv_left, vec_recv_right )
-                # An MPI barrier is needed here so that a single rank does not
-                # do two sends and receives before this exchange is completed.
-                self.mpi_comm.Barrier()
                 # Handle the received buffers
                 self.mpi_buffers.handle_vec_buffer(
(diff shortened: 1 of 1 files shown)


RemiLehe:
Nice! Thanks for this PR!
I also get a ~5% speedup on Lawrencium with this PR.

PR #219, opened 2018-03-04
[WIP] Add CUDA GPUDirect support for direct GPU MPI communication

MKirchen:
**Please merge PR #217 before!**

This PR introduces the support for leveraging the GPUDirect technology which allows to directly send CUDA device arrays between GPUs over MPI. In addition - depending on the cluster implementation - the feature can lead to a significant speed up for the GPU MPI communication in FPBIC.

If the environment variable `FBPIC_ENABLE_GPUDIRECT=1` is set, FBPIC will directly exchange the field information in the guard region between CUDA devices without copying the data to the CPU host first. For the particle exchange (which occurs only every `exchange_period` iterations) the data is still send to the CPU first.

Additional notes:
- I added a small section to the documentation that explains how to enable this feature
- The printing message will inform the user if GPUDirect is used if `verbose_level=2` is set

Three different cases have been benchmarked (average time per step) with the script attached below.
Setup: JURECA, 2x4 = 8xK80 on 2 nodes with 8 MPI processes. (MVAPICH2/2.2-GDR & CUDA 8)

1) Barrier() and old Wait() pattern (master):   94 ms
2) no Barrier() and new Wait() pattern (PR #217):   90 ms
3) GPUDirect (based on PR #217) (this PR):   82 ms

[fbpic_mpi_benchmark.py.zip](https://github.com/fbpic/fbpic/files/1778905/fbpic_mpi_benchmark.py.zip)

In order to test if GPUDirect is supported on a cluster, one can use the following example script using mpi4py and numba:

[gpudirect_numba.py.zip](https://github.com/fbpic/fbpic/files/1778907/gpudirect_numba.py.zip)

PR #226, opened 2018-03-15
Implement cuda-aware MPI communications

RemiLehe:
This pull request reimplements PR #219 (which was reverted by PR #225 because some tests were observed to fail, on machines that support GPUDirect).

### DIFF
--- a/docs/source/how_to_run.rst
+++ b/docs/source/how_to_run.rst
@@ -53,6 +53,23 @@ Python script: either ``lwfa_script.py`` or
 	export FBPIC_DISABLE_THREADING=1
 	python fbpic_script.py
 
+.. note::
+
+  When running on GPU with MPI domain decomposition, it is possible to enable
+  the CUDA GPUDirect technology. GPUDirect enables direct communication of
+  CUDA device arrays between GPUs over MPI without explicitly copying the data
+  to CPU first, resulting in reduced latencies and increased bandwidth. As this
+  feature requires a CUDA-aware MPI implementation that supports GPUDirect,
+  it is disabled by default and should be used with care.
+
+  To activate this feature, the user needs to set the following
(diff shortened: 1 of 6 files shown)


MKirchen:
The problem has been fixed. A cuda.synchronize() call was necessary before sending the MPI buffer arrays directly from GPU memory over MPI. The most likely reason for the problem is the asynchronous kernel execution in combination with the nonblocking MPI commands. It can happen that the non-blocking MPI send/recv calls access and send GPU memory that is still written by a GPU kernel that is executing asynchronously.

I only added a single cuda.synchronize() call directly after calling the GPU kernel that fills up the MPI buffers and before the actual send/recv over MPI. I figured that there is no cuda.synchronize() needed after the MPI exchange calls, because the processes MPI.Wait() for each other until the transfer is complete. Thus, the subsequent kernel that copies the exchanged buffer content to the field arrays will not start to execute too early.

Updated benchmark results ( using the script provided with PR #219 ):
Same setup: JURECA, 2x4 = 8xK80 on 2 nodes with 8 MPI processes

- no Barrier() and new Wait() pattern ( PR #217 ): 90 ms
- GPUDirect ( PR #219 ): 82 ms
- Fixed GPUDirect but disabled ( PR #226 ) (this PR): 80 ms
- Fixed GPUDirect enabled ( PR #226 ) (this PR): 72 ms
- Current master: 94 ms

I don't know why the execution is faster than the current master branch without GPUDirect (I tested on the exact same nodes with the same configuration)... Any ideas @RemiLehe ?

RemiLehe:
Great! Thanks a lot for fixing the bug!

Regarding the faster results with the current `dev` (compared to previous tests): could it be that the tests were initializing (and handling) zero-weight particles - which are scraped in the current `dev`?

PR #230, opened 2018-03-23
Clean up and generalize continuous injection

RemiLehe:
This is a major refactoring of the continuous injection / moving window. Sorry for the size of this PR! I'll try to explain it in details here.

With this PR, the information that is needed in order to continuously inject particles is now stored in a `ContinuousInjector` method. This includes:
- The density (`n`, `dens_func`) and number of particles in r and theta (`Npr`, `Npt`) which used to be stored in the `Particles` object.
- The positions that keep track of where the particles should be injected in the next iterations (`z_inject`, `z_end_plasma`, `v_end_plasma`) as well as the velocities of the injected particles (`ux_m`, `ux_th`, etc.), which used to be stored (with a single value for all species) in the `MovingWindow` object.
- A new attribute `dz_particles` which is needed in order to know the spacing between particles that are injected (when no particles are injected in the initial box). This replaces the attribute `p_nz` of the `Simulation` object.

These changes have several important consequences:
- Because `ux_m`, `ux_th`, etc. are not stored at the `MovingWindow` level, they do not need to be passed when initializing the moving window. (They used to be optional arguments.) These arguments are kept (for backward compatibility), but a `DeprecationWarning` will be issued if the user passes them. 
- Because this PR gets rid of a single `p_nz` for all species, we can now inject species with different number of particles per cell in z. This is now automatically tested in `test_continous_injection.py`.

In addition, I also cleaned up the way in which the particles are generated during continuous injection. We used to create *a new `Particles`* object at each `exchange_period`, and then add the generated positions/velocities of the corresponding macroparticles to the existing `Particles` object. This was done in order to reuse the code that generates evenly-spaced macroparticles in the `Particles.__init__` function. However, generating a new `Particles` object is heavy, and it imposed to do ugly stuff, like the flag `full_initialization=False` in the `make_ionizable` function.

In the current PR, the code that generates evenly-spaced macroparticles is separated in a dedicated function (`generate_evenlyspaced`), so that we do not need to create a new `Particles` object. This new function is called in the `Particles.__init__`, as well as in a new function `generate_continuously_injected_particles` which is called at each `exchange_period`. The flag `full_initialization` is removed.


fbpic/particles/utilities/utility_methods.py
```diff
@@ -102,54 +102,3 @@ def weights(x, invdx, offset, Nx, direction, shape_order):
 
     # Return the result
     return( i, S )
```
RemiLehe: This has been moved into the file that contains the `ContinuousInjector` object.

MKirchen:
Nice, thanks for implementing this!

PR #231, opened 2018-03-23
Close #187 Document ionization module and add an example script 

### LINKED ISSUE #187: Document ionization module and add an example script
We might want to have a simpler API for creating Particles for this... Even a small backward-compatible change would be good. I'll think about it.

RemiLehe:
This PR update and expands the documentation by:
- Encouraging the user to use the new function `add_new_species`
- Adding a section on the `Particles` object, which includes the documentation of the `make_ionizable` method. Note that the `__init__` method of the `Particles` object is **not** documented.
- Adding a new downloadable script for ionization simulation. Note that `add_new_species` makes this script much simpler than it would have been in the previous fbpic release. This script is now also automatically tested in `test_example_doc_scripts.py`. Note that `add_new_species` is now also used in the automated test `test_ionization.py`.

The modified documentation can be seen here: http://www.normalesup.org/~lehe/html_fbpic/

This PR also makes a number of arguments of the `Simulation` object optional (those that correspond to the creation of macroparticles at the beginning of the simulation). The defaults values are the same as of the `add_new_species` method. Omitting these arguments is useful/natural for scripts with ionization.

Finally, this PR also removes the messages of the type `Getting cashed ionization energy` that appear when running ionization simulations.

### DIFF
--- a/docs/source/api_reference/api_reference.rst
+++ b/docs/source/api_reference/api_reference.rst
@@ -8,6 +8,7 @@ the ``fbpic`` package, and which are used to set up a simulation.
    :maxdepth: 1
 
    simulation
+   particles
    diagnostics
    checkpoint_restart
    lpa_utilities/lpa_utilities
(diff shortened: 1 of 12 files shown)


MKirchen:
Great!

PR #232, opened 2018-03-26
Fix bunch initialization

RemiLehe:
This PR fixes a few bugs in the bunch initialization:

- for the function `add_elec_bunch` (i.e. uniform bunch), `p_zmin` and `p_zmax` were not properly transformed in boosted-frame simulations. This is fixed by taking into account the `beta` of the plasma inside `add_new_species`.

- for the function `add_elec_bunch_from_arrays`, the particles were selected in the local domain *before* they were Lorentz-transformed. Because the local domain boundaries are expressed in the boosted frame, this was incorrect. In order to fix this bug, I had to fix the function `boost_particles`, so that it can take arrays as inputs (instead of taking a `Particles` object)

In addition, the method `add_new_species` is now also used when creating the particles from arrays. More precisely, an empty species is created (no macroparticles), and then the arrays are reallocated, by using the function `reallocate_and_copy_old`, which is also used in Compton/ionization. This avoids code duplication, and also automatically handles particle tracking of the newly created particles. 
Note that I did have to modify the function `reallocate_and_copy_old` in order to detect if the arrays are currently on CPU or GPU. (This is because, even when `use_cuda` is True, `add_new_species` creates the particles on CPU)

### DIFF
--- a/fbpic/lpa_utils/boosted_frame.py
+++ b/fbpic/lpa_utils/boosted_frame.py
@@ -219,7 +219,7 @@ class BoostConverter( object ):
 
         return( boosted_frame_vars )
 
-    def boost_particles( self, particles ):
+    def boost_particle_arrays( self, x, y, z, ux, uy, uz, inv_gamma ):
         """
         Transforms particles to the boosted frame and propagates
         them to a fixed time t_boost = 0. without taking any electromagnetic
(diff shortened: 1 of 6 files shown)


MKirchen:
👍

PR #233, opened 2018-03-27
Initialize injection positions at the beginning of step

RemiLehe:
In the current `dev` branch, the positions that keep track of plasma injection are initialized when `set_moving_window` is called.

However, this is not ideal, since it relies on `set_moving_window` being called after all continuously-injected species are created. With `add_new_species`, users might inadvertently create new species after the moving window is set.

Therefore, this PR moves the initialization of these positions at the beginning of `step`. Note that, if `step` is called multiple times, these variables are not reinitialized (because of the `if self.z_inject is None` condition).

On the other hand, in the rare case where a user calls `step`, then does a restart, and then calls `step` again, these variables to need to be reinitialized. I added a new method `reset_injection_position` for this purpose, which gets called when doing a restart.

### DIFF
--- a/fbpic/boundaries/moving_window.py
+++ b/fbpic/boundaries/moving_window.py
@@ -15,25 +15,18 @@ class MovingWindow(object):
     """
     Class that contains the moving window's variables and methods
     """
-    def __init__( self, interp, comm, dt, ptcl, v, time ):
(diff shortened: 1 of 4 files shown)


MKirchen:
🙏

PR #234, opened 2018-03-28
Close #142 Do the `sum_reduce` only once per deposition (instead of once per species)

### LINKED ISSUE #142: Do the `sum_reduce` only once per deposition (instead of once per species)
The `sum_reduce` operation (on CPU) is expensive when using a large number of cores, and is being done for each particle species. One solution would be to modify the code so as to only perform this operation only once per deposition (just like the `divide by volume` operation)

RemiLehe:
In the current `dev` branch, the arrays `rho_global` and `J*_global` (which duplicate the charge/current arrays once per thread) are reallocated and summed for each species and at each iteration. This is obviously unnecessary and wasteful.

Therefore, in this PR:
- The arrays `rho_global` and `J*_global` are allocated only *once* (at initialization). And all species use the same arrays. As a consequence, these arrays are stored at the `Fields` level and not at the `Particles` level anymore. Also, because these arrays store all azimuthal modes together, I did not store them inside the `InterpolationGrid` object (which has one separate instance per azimuthal mode)
- The `sum_reduce` operation is done only once per iteration, after all species have deposited their charge/current.

The above brings a non-negligible speed-up in the case where multiple species are used. For instance, here are some benchmarks on a 2-core CPU (note that the speed up should also increase with the number of threads, because the `sum_reduce` typically takes a larger fraction of the computer time in this case):

**ionization_script.py (without IO, excluding compilation time)** (4 species):
- `dev`: 176 ms/iteration
- This PR: 156 ms/iteration

**lwfa_script.py (without IO, excluding compilation time)** (only one species):
- `dev`: 84 ms/iteration
- This PR: 81 ms/iteration (maybe because we don't need to reallocate the arrays at each iteration)

Note that the arrays `rho_global` and `J*_global` can take a lot of memory. Therefore, by default, they are not initialized (e.g. when doing direct deposition of the laser and space-charge calculations, we create a `Fields` object that spans across the *global domain*, but we only use it to do the spectral transforms, not to do any depositions, in these cases we do not want to create these large arrays). The `__init__` function of the `Simulation` object is the only place were the flag for the creation of these arrays is set to True (and then, only when running on CPU of course).


MKirchen:
👍

PR #236, opened 2018-03-30
Close #117 Each species should have its own `prefix_sum_shift`

### LINKED ISSUE #117: Each species should have its own `prefix_sum_shift`, `z_inject` and `z_end_plasma`
These different quantities are used by the moving window to inject particles. 

However, these values could in principle be different for each species. For instance, for species that are sorted more rarely (e.g. photons), `prefix_sum_shift` is different than that of species which are sorted frequently.

RemiLehe:
When shifting the moving window, we avoid having to sort the particles again (as long as the particles are not pushed by the particle pusher) by setting the `prefix_sum_shift` variable, which keeps track of the number of cells that the moving window moved, since the last time that the particles were sorted.

Currently, this is a single attribute for all particle species. However, this assumes that all species are always sorted at the same time. This is not necessarily the case, e.g. because some species could be involved in boosted-diagnostics (which involve particle sorting in order to extract particle slices) whereas others are not.

For this reason, `prefix_sum_shift` is now a per-species attribute.

I successfully ran the automated tests on GPU, for this PR.

### DIFF
--- a/fbpic/boundaries/moving_window.py
+++ b/fbpic/boundaries/moving_window.py
@@ -111,9 +111,10 @@ class MovingWindow(object):
 
         # Because the grids have just been shifted, there is a shift
         # in the cell indices that are used for the prefix sum.
-        if fld.use_cuda:
-            fld.prefix_sum_shift += n_move
-            # This quantity is reset to 0 whenever prefix_sum is recalculated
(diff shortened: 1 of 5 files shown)


MKirchen:
👍

PR #237, opened 2018-03-31
Close #213  Restart does not work for ionizable species

### LINKED ISSUE #213: Restart does not work for ionizable species
Right now, the ionization state of each ion is not loaded when doing a restart. This prevents a correct restart with ionization on.

RemiLehe:
The restart mechanism was incorrect for ionizable species, because the associated quantities (ionization level, weight multiplied by the level) were not set during the restart.

This PR fixes the issue, and adds an automated test that restart an ionization simulation and checks that the results are *close* to those of a simulation without restart. Note that, because of the randomness of the ionization (which is *not* controlled by the `numpy` random seed, but instead by some seed inside `numba`), the restarted simulation will not give the same results, and the tolerance with which the fields are checked must be adapted.

### DIFF
--- a/tests/test_example_docs_scripts.py
+++ b/tests/test_example_docs_scripts.py
@@ -155,7 +155,7 @@ def run_sim( script_name, n_MPI, checked_fields ):
     print('Checking particle ids...')
     start_time = time.time()
     for iteration in ts1.iterations:
-        pid, = ts1.get_particle(["id"], iteration=iteration)
+        pid, = ts1.get_particle(["id"], iteration=iteration, species="electrons")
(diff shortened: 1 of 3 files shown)


tests/test_example_docs_scripts.py
```diff
@@ -236,38 +263,6 @@ def test_parametric_sim_twoproc():
     # Suppress the temporary directory
     shutil.rmtree( temporary_dir )
 
-def test_ionization_script_twoproc():
```
RemiLehe: This test has been moved a few lines above, and reuses the function `run_sim`, so as to reduce the amount of code.

PR #239, opened 2018-03-31
Add various tips in the documentation

RemiLehe:
This adds various tips in the documentation, mainly based on feedback from users:

- Mentions that the number of particles should be increased when increasing the number of azimuthal modes.
- Mentions that boosted-frame simulations should include ions, and that it is good to check the convergence with lower gamma.
- Insists on using the parameter ``use_all_mpi_ranks`` when doing parameter scans (easily overlooked in the documentation)
- Mentions how to profile multi-GPU code
- Prevents the timestep from being too big in the example boosted-frame simulation.

The updated documentation can be seen here: http://www.normalesup.org/~lehe/html_fbpic/


MKirchen:
Thanks 🙏

PR #241, opened 2018-04-11
Correctly calculate the stencil extent in the Galilean scheme

soerenjalas:
With this PR the stencil extent for simulations using the Galilean scheme is calculated correctly. In contrast to lab frame simulation where the extent is calculated from the C and S coefficients in the field equation, in the Galilean scheme Theta^2*C and Theta^2*S need to be used.


RemiLehe:
Thanks for this PR @soerenjalas !!

PR #244, opened 2018-04-20
Add laser profile for flattened Gaussian laser

RemiLehe:
This adds a laser profile which is flattened in the near-field and has rings in the far-field. The advantage of this profile is that it has a fully analytical expression for any `z` and `t` (contrary to an exact flat-top profile).

In addition, the documentation for laser profiles has been slightly modified, in order to list the different available laser profiles, and have one page per profile, as this makes it more readable ; here is a preview: http://www.normalesup.org/~lehe/html_fbpic/api_reference/lpa_utilities/laser.html#laser-profiles

TODO

- [x] Add automated test

### DIFF
--- a/tests/test_flattenedgauss_laser.py
+++ b/tests/test_flattenedgauss_laser.py
@@ -1,0 +1,112 @@
+# Copyright 2016, FBPIC contributors
+# Authors: Remi Lehe, Manuel Kirchen
+# License: 3-Clause-BSD-LBNL
+"""
+This test file is part of FB-PIC (Fourier-Bessel Particle-In-Cell).
+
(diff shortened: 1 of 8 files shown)


MKirchen:
👍

tests/test_flattenedgauss_laser.py
```diff
@@ -0,0 +1,112 @@
...
+    return u
+
+
+def test_laser_periodic(show=False):
+    """
+    Function that is run by py.test, when doing `python setup.py test`
+    Test the propagation of a laser in a periodic box.
+    """
+    # Propagate the pulse in a single step
+    dt = Lprop*1./c
```
MKirchen: So cool ;)

PR #247, opened 2018-04-22
Simplify MPI code, and use separate buffers for E and B exchange

RemiLehe:
This PR is more of a suggestion ; feel free to reject it.

The aim is to reduce duplication in the code for the MPI exchanges. This is done (mainly) by using dictionaries that point to the different types of MPI buffers, instead of using `if` conditions which explicitly refer to the buffer in each case.

Also, there is now a separate buffer for each of the exchanges that are done (replace B, replace E, add J, add rho). In the future, this could be useful in order to make these exchanges asynchronous, and overlap them. There is no buffer anymore for the exchanges that are not done in the code (replacing a scalar).

I successfully tested this with 2 MPI, using the `test_linear_wakefield.py` with `n_order=32` in the following configurations:
- on CPU
- on GPU without GPU-direct
- on GPU with GPU-direct

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -483,224 +483,71 @@ class BoundaryCommunicator(object):
             of field exchange that is needed
         """
         # Only perform the exchange if there is more than 1 proc
-        if self.size > 1:
-            if fieldtype == 'E':
-                if method == 'replace':
-                    if gpudirect_enabled:
-                        vec_send_left = self.get_gpu_mpi_buffer(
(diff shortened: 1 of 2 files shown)


fbpic/boundaries/boundary_communicator.py
```diff
@@ -737,25 +584,6 @@ def exchange_domains( self, send_left, send_right, recv_left, recv_right ):
             req_rr.Wait()
             req_sr.Wait()
```
RemiLehe: Because this method did not use `self`, I made it a separate helper function.

MKirchen:
Great! I like it a lot!

PR #250, opened 2018-04-25
Implement laser envelope model in vacuum

PierrePelletier:
This is a first draft for implementing an envelope model for the laser, which would allow for a lower resolution along the z axis.
In this PR we modified heavily fields.py and its dependencies to allow for the handling of the envelope of the laser. We thus store the A envelope field, as well as the dtA field (derivative in time) and we are able to push them forward in time. For this, as the envelope is inherently complex, we also need to compute the negative azimuthal modes and so we store 2Nm - 1 modes in total. In order to do this, the InterpolationGrid and SpectralGrid objects have each been separated in two variations, Envelope and Field which contains everything the old grids used to hold.

The simulation is for now capable of simulating the propagation of the A envelope field in a vacuum. New methods have also been created to initialize the field created by a Gaussian laser as an envelope A field rather than a standard E field. There is also a test file added allowing to test the propagation of the envelope in a vacuum, currently only for m = 0.

The following has not yet been implemented:
-Any interactions with the plasma
-MPI or GPU use 
-Comoving approach (Note: the envelope model as of now is not valid for high boosts)
-Handling moving windows
-Boundary conditions
-Any further tests

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -117,6 +117,8 @@ class Fields(object) :
         self.n_order = n_order
         self.v_comoving = v_comoving
         self.use_galilean = use_galilean
+        self.zmin = zmin
+        self.zmax = zmax
 
         # Define wether or not to use the GPU
         self.use_cuda = use_cuda
@@ -172,26 +174,6 @@ class Fields(object) :
                                 V=self.v_comoving,
                                 use_galilean=self.use_galilean,
                                 use_cuda=self.use_cuda ) )
-                                
-        #Create the envelope interpolation grids for each modes
-        #The envelope modes range from -Nm + 1 to Nm - 1
-        self.envelope_interp = []
-        for m in range(2*Nm-1):
-            #Modes are listed in order: 0, 1, ..., Nm - 1, -Nm + 1, ..., -1
-            mode = self.get_mode(m)
-            self.envelope_interp.append(EnvelopeInterpolationGrid(
-                Nz, Nr, mode, zmin, zmax, rmax, use_cuda=self.use_cuda ) )
-               
-        #Create the envelope spectral grids for each modes
-        self.envelope_spect = []
-        for m in range(2*Nm-1):
-            #Modes are listed in order: 0, 1, ..., Nm - 1, -Nm + 1, ..., -1
-            mode = self.get_mode(m)
-            kr = 2*np.pi * self.trans[abs(mode)].dht0.get_nu()
-            self.envelope_spect.append( EnvelopeSpectralGrid( kz_modified, kr, mode,
-                kz_true, self.envelope_interp[m].dz, self.envelope_interp[m].dr,
-                use_cuda=self.use_cuda ) )
-        
 
         # Record flags that indicates whether, for the sources *in
         # spectral space*, the guard cells have been exchanged via MPI
@@ -213,6 +195,54 @@ class Fields(object) :
                     shape=(nthreads, self.Nm, self.Nz+4, self.Nr+4) )
             self.Jz_global = np.zeros( dtype=np.complex128,
                     shape=(nthreads, self.Nm, self.Nz+4, self.Nr+4) )
+             
+        # By default will not use the envelope model
+        self.use_envelope = False
+        
+    
+    def activate_envelope_model(self, k0):
+        """
+        Initializes anything needed for the envelope model
+        
+        Attribute use_envelope can be changed to true only there,
+        and thus guarantees the existence of all the attributes relevant
+        to the envelope model.
+        
+        Parameters
+        ----------
+        k0: float
+            Wavenumber of the beam represented by the envelope model
+            It is important to have only one well-defined wavelength
+            in this model     
+        """
+        
+        self.use_envelope = True
+        
+        #Create the envelope interpolation grids for each modes
+        #The envelope modes range from -Nm + 1 to Nm - 1
+        self.envelope_interp = []
+        for m in range(2 * self.Nm - 1):
+            #Modes are listed in order: 0, 1, ..., Nm - 1, -Nm + 1, ..., -1
+            mode = self.get_mode(m)
+            self.envelope_interp.append(EnvelopeInterpolationGrid(
+                self.Nz, self.Nr, mode, self.zmin, self.zmax, self.rmax, use_cuda=self.use_cuda ) )
+               
+        #Create the envelope spectral grids for each modes
+        self.envelope_spect = []
+        dz = (self.zmax-self.zmin)/self.Nz
+        kz_true = 2*np.pi* np.fft.fftfreq( self.Nz, dz )
+        kz_modified = get_modified_k( kz_true, self.n_order, dz )
+        for m in range(2 * self.Nm - 1):
+            #Modes are listed in order: 0, 1, ..., Nm - 1, -Nm + 1, ..., -1
+            mode = self.get_mode(m)
+            kr = 2*np.pi * self.trans[abs(mode)].dht0.get_nu()
+            self.envelope_spect.append( EnvelopeSpectralGrid( kz_modified, kr, mode,
+                kz_true, self.envelope_interp[m].dz, self.envelope_interp[m].dr,
+                use_cuda=self.use_cuda ) )
+        
+        # Create the psatd coefficients relevant only to the envelope model for each positive mode
+        for m in range(self.Nm):        
+            self.psatd[m].compute_envelope_coefs(self.spect[m].kz, self.spect[m].kr, m, self.dt, self.Nz, self.Nr, k0)
 
 
     def send_fields_to_gpu( self ):
@@ -269,9 +299,9 @@ class Fields(object) :
         for m in range(self.Nm) :
             self.spect[m].push_eb_with( self.psatd[m], use_true_rho )
 
-            # Placeholder for determining if one should use the envelope model
-        use_envelope_model_bool = True
-        if (use_envelope_model_bool):
+        # Check if the envelope model is used then 
+        # push each azimuthal mode individually
+        if (self.use_envelope):
             for m in range(2*self.Nm - 1) :
                 mode = self.get_mode(m)
                 self.envelope_spect[m].push_envelope_with(self.psatd[abs(mode)])
@@ -323,7 +353,7 @@ class Fields(object) :
         ---------
         fieldtype :
             A string which represents the kind of field to transform
-            (either 'E', 'B', 'J', 'rho_next', 'rho_prev’, ‘A’)
+            (either 'E', 'B', 'J', 'rho_next', 'rho_prev', 'A')
         """
         # Use the appropriate transformation depending on the fieldtype.
         if fieldtype == 'E' :
@@ -356,7 +386,7 @@ class Fields(object) :
                 spectral_rho = getattr( self.spect[m], fieldtype )
                 self.trans[m].interp2spect_scal(
                     self.interp[m].rho, spectral_rho )
-        elif fieldtype == 'A':
+        elif fieldtype == 'A' and self.use_envelope:
             # Transform each azimuthal grid individually
             for m in range(2*self.Nm - 1) :
                 mode = self.get_mode(m)
@@ -413,7 +443,7 @@ class Fields(object) :
             for m in range(self.Nm) :
                 self.trans[m].spect2interp_scal(
                     self.spect[m].rho_prev, self.interp[m].rho )
-        elif fieldtype == 'A' :
+        elif fieldtype == 'A' and self.use_envelope:
             # Transform each azimuthal grid individually
             for m in range(2*self.Nm - 1) :
                 mode = self.get_mode(m)
@@ -421,6 +451,7 @@ class Fields(object) :
                     self.envelope_spect[m].A, self.envelope_interp[m].A )
                 self.trans[abs(mode)].spect2interp_scal(
                     self.envelope_spect[m].dtA, self.envelope_interp[m].dtA )
+                
         else :
             raise ValueError( 'Invalid string for fieldtype: %s' %fieldtype )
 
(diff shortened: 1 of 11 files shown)


RemiLehe:
From offline discussions with @PierrePelletier: this PR will be reopened after implementing a mechanism to initialize the laser envelope only when needed. (right now the envelope is used by default)

PierrePelletier:
Reopening this PR since the mechanism to initialize the laser envelope has been implemented.

This mechanism work by setting a flag in the Field object when you call the method adding the laser envelope to the mesh.
On top of that, the code has now been tested for mode m = 1 with a LaguerreGaussian profile

RemiLehe:
Great! Thanks for reopening the PR. I'll start reviewing it tomorrow.

fbpic/fields/fields.py
```diff
@@ -446,7 +532,7 @@ def partial_interp2spect(self, fieldtype) :
         ---------
         fieldtype :
             A string which represents the kind of field to transform
-            (either 'E', 'B', 'J', 'rho_next', 'rho_prev')
+            (either 'E', 'B', 'J', 'rho_next', 'rho_prev’, ‘A’)
```
RemiLehe: Could you replace the back ticks here (`) by regular ticks (')?
I think that this is what is making the Python2 tests fail.

fbpic/fields/fields.py
```diff
@@ -571,3 +673,15 @@ def divide_by_volume( self, fieldtype ) :
         """
         for m in range(self.Nm):
             self.interp[m].divide_by_volume( fieldtype )
+
```
RemiLehe: If we create a list `self.envelope_mode_numbers`, then this method can be suppressed.

fbpic/fields/interpolation_grid.py
```diff
@@ -70,18 +70,6 @@ def __init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False ) :
         # Note: No Verboncoeur-type correction required
         self.invvol = 1./vol
```
RemiLehe: Could you update the docstring of the `InterpolationGrid` to say that this is a base class, which is used for the `FieldInterpolationGrid` and `EnvelopeInterpolationGrid`?

fbpic/fields/fields.py
```diff
@@ -496,12 +590,16 @@ def erase(self, fieldtype ) :
         ---------
         fieldtype : string
             A string which represents the kind of field to be erased
-            (either 'E', 'B', 'J', 'rho')
+            (either 'E', 'B', 'J', 'rho' or 'A')
```
RemiLehe: For now, since we do not need the `erase` method, I would recommend reverting it to its previous version (without A)

fbpic/fields/fields.py
```diff
@@ -549,11 +647,15 @@ def filter_spect( self, fieldtype ) :
         ---------
         fieldtype : string
             A string which represents the kind of field to be filtered
-            (either 'E', 'B', 'J', 'rho_next' or 'rho_prev')
+            (either 'E', 'B', 'J', 'rho_next', 'rho_prev' or 'A')
```
RemiLehe: For now, since we do not need the `filter` method yet, I would recommend to revert it to its previous version (without A)

fbpic/fields/interpolation_grid.py
```diff
@@ -218,3 +229,24 @@ def divide_by_volume( self, fieldtype ) :
...
+class EnvelopeInterpolationGrid(InterpolationGrid):
+    
+    def __init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False ) :
+        
+        InterpolationGrid.__init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False)
+        
+        # Allocate the fields arrays
+        self.A = np.zeros( (Nz, Nr), dtype='complex' )
+        self.dtA = np.zeros( (Nz, Nr), dtype='complex' )
+
```
RemiLehe: For now, the `erase` method is not needed, and so I would recommend to just remove it for the moment.

fbpic/fields/numba_methods.py
```diff
@@ -139,7 +139,26 @@ def numba_push_eb_standard( Ep, Em, Ez, Bp, Bm, Bz, Jp, Jm, Jz,
                 + j_coef[iz, ir]*( 1.j*kr[iz, ir]*Jp[iz, ir] \
                             + 1.j*kr[iz, ir]*Jm[iz, ir] )
 
+
     return
+    
+@njit_parallel
+def numba_push_envelope_standard(A, dtA, w2_square, invw_tot, S_env, C_env,
+                            sinc_env, A_coef, Nz, Nr):
```
RemiLehe: Could you add a docstring for this function? (on the same model as for the `push_eb_standard`)

fbpic/fields/psatd_coefs.py
```diff
@@ -173,3 +173,37 @@ def __init__( self, kz, kr, m, dt, Nz, Nr, V=None,
...
+        dt : float
+            The timestep of the simulation
+        
+        k0: float
+            Wavenumber of the beam represented by the envelope model
+        """
+        
+        # Calculate all necessary coefficients for propagation of A field
+        w_laser = c * k0
+        self.w2_square = c**2 * (kr**2 + kz**2 + 2 * kz * w_laser/c)
```
RemiLehe: Could you just use `k0` instead of `w_laser/c` in the above expression?

fbpic/fields/spectral_grid.py
```diff
@@ -410,3 +432,48 @@ def filter(self, fieldtype) :
...
+            ps : PsatdCoeffs object
+                psatd object corresponding to the same m mode
+            """
+            assert (ps.V is None or ps.V == 0)
+            assert( abs(self.m) == ps.m )
+        
+            numba_push_envelope_standard(self.A, self.dtA, ps.w2_square, ps.invw_tot, ps.S_env,
+                ps.C_env, ps.sinc_env, ps.A_coef, self.Nz, self.Nr)
+
+
```
RemiLehe: For now, I think that the `filter` function is not needed, and so it could be removed.

fbpic/fields/psatd_coefs.py
```diff
@@ -173,3 +173,37 @@ def __init__( self, kz, kr, m, dt, Nz, Nr, V=None,
...
+        w_tot = np.sqrt( (w_laser + c * kz)**2 + c**2 * kr**2)
```
RemiLehe: Instead of `invw_tot` and `S_env` could we actually store their product (since this is what is used in the actual push function)?
e.g.
```
self.S_env_over_w = np.sin(w_tot*dt) / np.where( w_tot == 0, 1, w_tot )
self.S_env_over_w[ w_tot==0 ] = dt
```
This would have the advantage of computing the correct value for the limit `w_tot = 0`.

Of the course the `push_envelope` function would have to be modified.

tests/unautomated/test_laser_envelope.py
```diff
@@ -0,0 +1,396 @@
...
+# Parameters
+# ----------
+# (See the documentation of the function propagate_pulse
+# below for their definition)
+show = False # Whether to show the plots, and check them manually
+
+use_cuda = False
+
+# Simulation box
+Nz = 400
```
RemiLehe: Actually, the envelope solver should work also when the laser wavelength is badly resolved ; therefore, I am wondering if using a small `Nz` (e.g. `Nz = 80`) would still work for this test. 
If yes, it would be great to modify this parameter accordingly.

tests/unautomated/test_laser_envelope.py
```diff
@@ -0,0 +1,396 @@
...
+zmax = 10.e-6
+Nr = 25
+Lr = 40.e-6
+Nm = 2
+n_order = -1
+# Laser pulse
+w0 = 4.e-6
+ctau = 5.e-6
+k0 = 2*np.pi/0.8e-6
+E0 = 1.
```
RemiLehe: Maybe we should replace E0 by a0 in this test.

tests/unautomated/test_laser_envelope.py
```diff
@@ -0,0 +1,396 @@
...
+    fld : Fields object from fbpic
+
+    m : int, optional
+       The index of the mode to be fitted
+    """
+    # Integrate the laser oscillations longitudinally
+    dz = fld.interp[0].dz
+    laser_profile = np.sqrt(dz*(abs( fld.envelope_interp[m].A )**2).sum(axis=0)) 
+    # Renormalize so that this gives the peak of the Gaussian
+    laser_profile *= 2.**(-3./4)/( np.pi**(1./4) * ctau**(1./2) )
```
RemiLehe: Actually, the aim of the 3 above lines was originally to evaluate the amplitude of the laser envelope, in the case where the available data is an oscillating electric field.

However, here, the available data is the laser envelope directly. So we could replace these 3 lines by simply
```
laser_envelope = abs( fld.envelope_interp[m].A )
```

tests/unautomated/test_laser_envelope.py
```diff
@@ -0,0 +1,396 @@
...
+                    exchange_period = 1, use_galilean=use_galilean )
+    # Remove the particles
+    sim.ptcl = []
+    # Set the moving window object
+    if v_window !=0:
+        sim.set_moving_window( v=v_window )
+
+    # Initialize the laser fields
+    z0 = (zmax+zmin)/2
+    init_fields( sim, w0, ctau, k0, z0, zf, E0, m )
```
RemiLehe: We could again use `a0` instead of `E0` here.

tests/unautomated/test_laser_envelope.py
```diff
@@ -0,0 +1,396 @@
...
+    else:
+        assert np.allclose( w, w_analytic, rtol=rtol )
+        assert np.allclose( A, A_analytic, rtol=5.e-3 )
+        print('The simulation results agree with the theory to %e.' %rtol)
+
+    # Return a dictionary of the results
+    return( { 'A' : A, 'w' : w, 'fld' : sim.fld } )
+
+
+def init_fields( sim, w, ctau, k0, z0, zf, E0, m=1 ) :
```
RemiLehe: We could again use `a0` instead of `E0` here.

fbpic/lpa_utils/laser/direct_envelope_injection.py
```diff
@@ -0,0 +1,145 @@
...
+    # Check that no other laser simulation has been already added
+    if sim.fld.use_envelope:
+        raise ValueError("Another laser profile has already been added")
+    sim.fld.activate_envelope_model(laser_profile.k0)
+
+    # Get the local azimuthally-decomposed laser fields A and dtA on each proc
+    laser_A, laser_dtA = get_laser_A_dtA( sim, laser_profile, boost )
+    for m in range(2 * sim.fld.Nm - 1):
+        sim.fld.envelope_interp[m].A[:,:] = laser_A[:,:,m]
+        sim.fld.envelope_interp[m].dtA[:,:] = laser_dtA[:,:,m]
```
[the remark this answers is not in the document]
PierrePelletier: Oh right thank you for that I was mistaken on the number of modes that would be calculated. That also explains the troubles I had when testing for m = -1. 
But then why do we use ntheta = 2 * Nm when we only study 2* Nm -1 modes (even if studied simultaneously for the real fields)?

tests/unautomated/test_laser_envelope.py
PierrePelletier: I'm pretty sure this method wouldn't work since we actually need to make the z dependency disappear in the array.
I just tried evaluating the envelope at the estimated center of the beam in the z direction then fit the curve ( something like laser_profile = abs(fld.envelope_interp[m].A[izref]) with izref well chosen) which seems to work pretty well

fbpic/fields/interpolation_grid.py
```diff
@@ -99,6 +87,42 @@ def r(self):
...
+    
+    def __init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False ) :
+        """
+        Initialize a 'FieldInterpolationGrid' object
+
+        See the docstring of the parent class 'InterpolationGrid' 
+        for the meaning of the different parameters.
+        """
+        
+        InterpolationGrid.__init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False)
```
RemiLehe: Could you pass `use_cuda=use_cuda` instead of `use_cuda=False`?

fbpic/fields/interpolation_grid.py
```diff
@@ -218,3 +236,24 @@ def divide_by_volume( self, fieldtype ) :
...
+    """
+    Contains the coordinates and envelope of the spatial grid.
+    
+    Main attributes:
+    - z,r : 1darrays containing the positions of the grid
+    - A, dtA:
+      2darrays containing the envelope amplitude.
+    """
+    
+    def __init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False ) :
```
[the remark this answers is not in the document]
RemiLehe: Could you also pass `use_cuda=use_cuda` here?

fbpic/fields/interpolation_grid.py
```diff
@@ -218,3 +242,30 @@ def divide_by_volume( self, fieldtype ) :
...
+        """
+        Initialize a 'EnvelopeInterpolationGrid' object
+
+        See the docstring of the parent class 'InterpolationGrid' 
+        for the meaning of the different parameters.
+        """
+        
+        InterpolationGrid.__init__(self, Nz, Nr, m, zmin, zmax, rmax, use_cuda=False)
```
RemiLehe: Could you also pass `use_cuda=use_cuda` here?

fbpic/fields/spectral_grid.py
```diff
@@ -112,10 +98,62 @@ def __init__(self, kz_modified, kr, m, kz_true, dz, dr,
...
+    def __init__(self, kz_modified, kr, m, kz_true, dz, dr,
+                        current_correction, use_cuda=False ) :
+        """
+        Initialize a 'FieldSpectralGrid' object
+
+        See the docstring of the parent class 'SpectralGrid' 
+        for the meaning of the different parameters.
+        """
+        SpectralGrid.__init__(self, kz_modified, kr, m, kz_true, dz, dr,
+                        use_cuda=False )
```
RemiLehe: Same comment on use_cuda

fbpic/fields/spectral_grid.py
```diff
@@ -410,3 +449,54 @@ def filter(self, fieldtype) :
...
+                        use_cuda=False ) :
+        """
+        Initialize a 'EnvelopeSpectralGrid' object
+
+        See the docstring of the parent class 'SpectralGrid' 
+        for the meaning of the different parameters.
+        """
+        
+        SpectralGrid.__init__(self, kz_modified, kr, m, kz_true, dz, dr,
+                        use_cuda=False )
```
RemiLehe: Same comment on use_cuda

PR #255, opened 2018-05-03
GPU implementation for pushing the laser envelope

PierrePelletier:
This is a continuation of PR 250, thus it should be merged only after having merged the PR250.
Here we continued developing the envelope model by adding the possibility of using cuda and GPU to push the fields and transform them between Interpolation and Spectral grids.
The code is very similar to the one used for the fields.

The code still doesn't include any of the following features:
-Any interactions with the plasma
-MPI use
-Comoving approach (Note: the envelope model as of now is not valid for high boosts)
-Handling moving windows
-Boundary conditions

### DIFF
--- a/fbpic/fields/spectral_grid.py
+++ b/fbpic/fields/spectral_grid.py
@@ -125,11 +125,7 @@ class FieldSpectralGrid(SpectralGrid):
         for the meaning of the different parameters.
         """
         SpectralGrid.__init__(self, kz_modified, kr, m, kz_true, dz, dr,
-<<<<<<< HEAD
-                        use_cuda=False )
-=======
                         use_cuda=use_cuda )
->>>>>>> merged_envelope
 
         Nr, Nz = self.Nr, self.Nz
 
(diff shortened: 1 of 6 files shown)


fbpic/fields/spectral_grid.py
```diff
@@ -497,6 +498,39 @@ def push_envelope_with(self, ps):
...
+            # Obtain the cuda grid
+            dim_grid, dim_block = cuda_tpb_bpg_2d( self.Nz, self.Nr)
+            # Push the fields on the GPU
+
+            cuda_push_envelope_standard[dim_grid, dim_block](self.A, self.dtA,
+                                        ps.w2_square, ps.S_env_over_w,
+                                        ps.C_env, ps.w_laser, ps.A_coef,
+                                        self.Nz, self.Nr )
+
+
```
RemiLehe: Again, could you use one blank line instead of 2?

RemiLehe:
Thanks for this PR @PierrePelletier!
I had a few minor styling comments (see inline comments).
One more important comment: the envelope coefficients (`C_env`, `A_coeff`) need to also be copy to the GPU, when `use_cuda` is `True` (see similar code for the regular coefficients, in the `PsatdCoeff` class).

PR #256, opened 2018-05-03
Implements moving windows for the envelope

PierrePelletier:
This PR follows the PR 255 since it uses the GPU methods added there.
It allows for the use of a moving window while using the envelope approximation, with both normal and GPU computations. A test for this has also been added in test_laser_envelope.py

Of note: The accuracy of test_laser_envelope.py in regards to the amplitude (for moving_window and periodic) was inconsistent with Nz (even when increasing Nz) and often got beyond 5e-3, I increased it to 6e-3 for now to pass the tests.


RemiLehe:
Great! Thanks for these changes!

PR #257, opened 2018-05-04
Close #253 Implement donut-like Laguerre-Gauss modes

### LINKED ISSUE #253: Clarify the definition of the LaguerreGauss modes
The documentation should emphasize that these laser pulses do not have a donut-like intensity pattern. We could add a donut-like Laguerre-Gauss profile later.

RemiLehe:
This pull request adds a new type of laser profile: the donut-like Laguerre-Gauss modes.
These modes have a cork-screw phase pattern and a donut-like intensity pattern.

While these modes can in principle be constructed from the regular Laguerre-Gauss modes, it is somewhat error-prone (esp. for large m). And in practice, this is a common enough laser profile that several users might want to use it.

This PR also adds a few checks for the regular Laguerre-Gauss modes, so as to prevent users from using negative`m`.

### DIFF
--- a/docs/source/api_reference/lpa_utilities/laser.rst
+++ b/docs/source/api_reference/lpa_utilities/laser.rst
@@ -56,6 +56,7 @@ own custom laser profiles.
 
     laser_profiles/gaussian
     laser_profiles/laguerre
+    laser_profiles/donut_laguerre
     laser_profiles/flattened
 
 Combining (summing) laser profiles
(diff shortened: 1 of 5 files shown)


RemiLehe:
For clarification, here are 3D picture of the Laguerre-Gauss and donut-like Laguerre-Gauss profiles, for m=0.

laguerre

donut

MKirchen:
Cool

PR #258, opened 2018-05-04
Envelope diagnostics

PierrePelletier:
This PR is based on the PR 250 but is independent of PR 255 and 256.
It adds to the field diagnostics the possibility to output the real and imaginary parts of A and dtA each considered a scalar fields as with rho. This is done this way because the fields diagnostics were intended to output only real solutions.


fbpic/openpmd_diag/data_dict.py
```diff
@@ -15,6 +15,8 @@
     "J" : np.array([-3., 1., 0., 1., 0., 0., 0.]),
     "E" : np.array([ 1., 1.,-3.,-1., 0., 0., 0.]),
     "B" : np.array([ 0., 1.,-2.,-1., 0., 0., 0.]),
+    "A" : np.array([ 1., 1.,-2.,-1., 0., 0., 0.]),
+    "dtA" : np.array([ 1., 1.,-3.,-1., 0., 0., 0.]),
```
RemiLehe: In fact, A (as stored in the code, and written in the file) is dimensionless, unlike the actual vector potential. (Unless I misunderstood your implementation)

Therefore, I would say that this should be:
```
    "A" : np.array([ 0., 0., 0., 0., 0., 0., 0.]),
    "dtA" : np.array([ 0., 0., -1., 0., 0., 0., 0.]),
```
Do you agree?

fbpic/openpmd_diag/field_diag.py
```diff
@@ -163,21 +165,69 @@ def write_dataset( self, field_grp, path, quantity ) :
...
+
+        path : string
+            The relative path where to write the dataset, in field_grp
+
+        quantity : string
+            Describes which envelope is being written.
+            (Either A or dtA)
+        """
+        path_real = "%s_real" %(path)
+        path_imag = "%s_imag" %(path)
```
RemiLehe: I know that this function is always called with `path` equal to `quantity`. 
However, could you maybe do `%(quantity)` instead as `%(path)`, as it is maybe more intuitive?

fbpic/openpmd_diag/field_diag.py
```diff
@@ -163,21 +165,69 @@ def write_dataset( self, field_grp, path, quantity ) :
...
+                # There is a factor 2 here so as to comply with the convention
+                # in Lifschitz et al., which is also the convention
+                # adopted in Warp Circ
+                dset_real[2*m-1,:,:] = 2*real_mode[:,:].real
+                dset_real[2*m,:,:] = 2*real_mode[:,:].imag
+                dset_imag[2*m-1,:,:] = 2*imag_mode[:,:].real
+                dset_imag[2*m,:,:] = 2*imag_mode[:,:].imag
+
+
+    def get_dataset( self, quantity, m, envelope = False ):
```
RemiLehe: Instead of adding the optional envelope argument, could simply do a check on quantity inside this function?
```
if quantity in ['A', 'dtA']:
    data_one_proc = getattr( self.fld.envelope_interp[m], quantity )
else:
    data_one_proc = getattr( self.fld.interp[m], quantity )

File: fbpic/openpmd_diag/field_diag.py [hunk 4]
```diff
@@ -163,21 +165,69 @@ def write_dataset( self, field_grp, path, quantity ) :
...
+        """
+        path_real = "%s_real" %(path)
+        path_imag = "%s_imag" %(path)
+        if field_grp is not None:
+            dset_real = field_grp[path_real]
+            dset_imag = field_grp[path_imag]
+        else:
+            dset_real = None
+            dset_imag = None
+
```
RemiLehe: Could you add some comments that explain what is being done here. For instance
```
# The envelope (`A`) is a complex field ; however openPMD can only save real fields
# (and represents them as an azimuthal decomposition, involving complex coefficients)
# Therefore, here we need to save the real part and imaginary part of A separately, 
# and we need to reconstruct their azimuthal decomposition:
# (A_{real})_m = (A_m + A_{-m}^*)/2    (A_{imaginary})_m = (A_m - A_{-m}^*)/(2i)
# Note that for m=0, this simply gives
# (A_{real})_0 = Re[ A_m ]        (A_{imaginary})_0 = Im[ A_m ]
```

fbpic/openpmd_diag/field_diag.py
```diff
@@ -163,21 +165,69 @@ def write_dataset( self, field_grp, path, quantity ) :
...
+        if self.rank == 0:
+            mode0 = mode0.T
+            dset_real[0,:,:] = mode0[:,:].real
+            dset_imag[0,:,:] = mode0[:,:].imag
+
+        for m in range(1,self.fld.Nm):
+            modep = self.get_dataset( quantity, m, envelope = True)
+            modem = self.get_dataset( quantity, -m, envelope = True)
+            real_mode = 0.5 * (modep + modem.conjugate())
+            imag_mode = -0.5 * 1j * (modep - modem.conjugate())
```
RemiLehe: Could you include the 2 above lines inside the `if self.rank == 0` condition.
This is because, for MPI simulations, `get_dataset` returns `None` for all ranks, except rank 0.

RemiLehe: Also, `-0.5 * 1j *` could be contracted into `-0.5j *` which might actually be more efficient.

fbpic/openpmd_diag/field_diag.py
```diff
@@ -277,7 +334,19 @@ def create_file_empty_meshes( self, fullpath, iteration,
                     # Setup the record to which they belong
                     self.setup_openpmd_mesh_record(
                         field_grp[fieldtype], fieldtype, dz, zmin )
-
+                elif fieldtype in ["A", "dtA"]:
```
RemiLehe: Could you add just a quick comment to explain here:
```
# The envelope field is complex, but openPMD supports only real fields
# Therefore, we need to save the real part and imaginary part separately.
```

PR #263, opened 2018-05-18
New method of pushing the 'a' field

PierrePelletier:
This PR main objective is to switch the scheme we use to push the a field forward in time. Instead of using the 'A' and 'dtA' fields at time n to obtain 'A' at time (n+1), we now use 'a' at time n-1 (named a_old) and at time n to obtain a at time (n+1). As such, dtA has been removed and different methods and formulas have been adapted. This choice was made with Remi Lehe offline in order to be able to incorporate the push due to the 'a' field better in the PIC loop.

On top of that, the 'A' field has been renamed 'a' in order to better show the fact that it is a dimensionless scalar.

Lastly, we have noted that the results of the simulations in periodic boundaries have deteriorated and we are currently investigating this issue.


tests/test_laser_envelope.py
```diff
@@ -45,9 +45,9 @@
 use_cuda = True
 
 # Simulation box
-Nz = 50
-zmin = -10.e-6
-zmax = 10.e-6
+Nz = 200
+zmin = -20.e-6
+zmax = 20.e-6
```
RemiLehe: Could you keep the previous `Nz` and `zmin`/`zmax`, and then reduce the time step for the test with the periodic box (basically use the same time step as for the moving window)?
This would help use remember that there is a problem for large timesteps...

RemiLehe:
Thanks for this PR! 
I had a few comments above. Would you have time to have a look?

RemiLehe:
As discussed with @PierrePelletier offline: this new discretization requires a higher resolution in z for the standard vacuum propagation tests. This is somewhat surprising and will be investigated further in the future ; in the end, we might need to switch back to the previous discretization.

PR #264, opened 2018-05-18
New flag for the envelope model

PierrePelletier:
This PR follows directly the PR 263 and changed the way we initialize the flag for the envelope model. Now it is a parameter directly passed for the initialization of the Simulation object. The add_laser method now only sends the relevant k_0 to the Fields object for the computation of PSATD coefs. Finally, there is a new flag noting if we have received the k_0 of the laser yet in the Fields object.


RemiLehe:
Thanks for this PR! 
I made some small modifications ; mainly, renaming `lambda0` to `lambda_envelope` to be more specific.

PR #268, opened 2018-05-25
Add the possibility to use several passes of smoothing, and a compensator.

RemiLehe:
This is more of a suggestion: it seems that it would be useful to have the possibility to add more smoothing, and maybe use a compensator. This is definitely a bit more "expert-mode"...

In addition, this PR reduces the size of the arrays for smoothing (by using two 1darrays, instead of a 2d array), and creates numba functions for smoothing on the CPU.


MKirchen:
Whats the status of this PR?

MKirchen:
Should we merge this?

RemiLehe:
As discussed off-line with @MKirchen, I will merge this now, and we can do another PR later to increase the guard region. (as the current number of guard cells should be enough for most purposes)

PR #271, opened 2018-06-01
Add a way to compute and gather the gradient of the A field

PierrePelletier:
Based on PR 264, this PR adds different methods to compute, transform and gather the gradient of the A field on the particles.  Indeed, it is needed to implement later the ponderomotive force, the foundation of the envelope model.
The gradients are calculated in the spectral space as it is easier (a simple multiplication does it),then are transformed back to interpolation space and finally we gather the A field and the gradient on the particles to obtain the gradient of A to the square.


fbpic/particles/particles.py
```diff
@@ -180,6 +186,13 @@ def dens_func( z, r ) ...
         self.By = np.zeros( Ntot )
 
         self.use_envelope = use_envelope
+        if self.use_envelope:
+            self.a2 = np.zeros( Ntot )
+            self.grad_a2_z = np.zeros( Ntot )
+            self.grad_a2_x = np.zeros( Ntot )
+            self.grad_a2_y = np.zeros( Ntot )
```
RemiLehe: Could you add `dtype='complex'` here? This because these buffers are first used in order to gather the complex a field.

RemiLehe: OK, based on off-line discussions: this actually not needed because of recent changes in the way that the gather functions were written.

RemiLehe:
Thanks for this PR @PierrePelletier !
I noticed that the new functions (using the tuples) were implemented on the branch `vay_pusher`. Therefore, I ported these functions into the branch `grad_a`, with [this commit](https://github.com/fbpic/fbpic/pull/271/commits/8de4ffc2bbffa23dd9b136d53d03237f6dca0c2e), which I pushed, so as to update the current PR.

Could you check that you agree with this change? Thanks.

PierrePelletier:
This commit seems completely adequate, thank you and sorry again for misplacing the changes in the wrong branch.

RemiLehe:
No worries! I'll have a second pass at reviewing this PR this evening.

fbpic/particles/gathering/cuda_methods.py
```diff
@@ -192,6 +199,169 @@ def gather_field_gpu_linear(x, y, z,
...
+        for m in m_tuple:
+            # Add contribution from mode m
+            a_m = a_tuple[m]
+            grad_a_r_m = grad_a_r_tuple[m]
+            grad_a_t_m = grad_a_t_tuple[m]
+            grad_a_z_m = grad_a_z_tuple[m]
+            # Calculate azimuthal complex factor
+            exptheta_m = 1.
+            for _ in range(m):
+                exptheta_m *= (cos - 1.j*sin)
```
RemiLehe: It seems that the above would not work for negative `m`. (This only impacts the GPU implementation: the CPU implementation uses a different expression.)

I would suggest replacing it with:
```
for _ in range(abs(m)):
    exptheta_m *= (cos - 1.j*sin)
if m < 0:
   exptheta_m = exp_theta.conjugate()
```
(This is a trick to avoid negative powers - i.e. divisions - which are more time-consuming than multiplications.)

fbpic/particles/gathering/cuda_methods.py
```diff
@@ -335,3 +505,138 @@ def gather_field_gpu_cubic(x, y, z,
...
```
RemiLehe: Same comment: this could be replaced by
```
for _ in range(abs(m)):
    exptheta_m *= (cos - 1.j*sin)
if m < 0:
   exptheta_m = exp_theta.conjugate()
```

fbpic/particles/particles.py
```diff
@@ -634,7 +652,7 @@ def push_x( self, dt, x_push=1., y_push=1., z_push=1. ) :
                 self.inv_gamma, self.Ntot,
                 dt, x_push, y_push, z_push )
 
-    def gather( self, grid ) :
+    def gather( self, grid, envelope_grid = None ) :
```
RemiLehe: I think that the argument `envelope_grid` could be removed. (It probably comes from a previous implementation of the gathering.)

PR #275, opened 2018-06-07
Enable restart for arbitrary modes

soerenjalas:
So far it was not possible to load checkpoints with more than 2 modes. This tiny PR changes that. As a quick test I simulated an off-axis beam with Nm = 4. The plot below shows the Ex field in mode 3 before and after restarting.
restart


RemiLehe:
Thanks a lot for fixing this @soerenjalas !

PR #276, opened 2018-06-09
Thread the function that erase charge/current buffers

RemiLehe:
I recently observed that, when using a high number of threads on CPU (e.g. 16 in the case I looked at), the part of the code that erases the duplicated buffers for charge/current can take a significant amount of time.

This is certainly due to the fact that this part is written in numpy, and is not threaded. This PR solves this problem by threading this particular operation, using the usual numba synthax.

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -8,7 +8,7 @@ It defines the high-level Fields class.
 import warnings
 import numpy as np
 from fbpic.utils.threading import nthreads
-from .numba_methods import sum_reduce_2d_array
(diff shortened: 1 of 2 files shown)


RemiLehe:
I just confirmed this by profiling a production-type simulation, using 16 threads per proc.
In this case `erase` went from being the 4th most expensive function (after `gather`, `deposit_J`, `deposit_rho`) to being 6th most expensive function (takes about the same time as `sum_reduce_2d_array`, which also operates on the threading buffers)

PR #277, opened 2018-06-23
Allow the user to specify the temperature, through `add_new_species`

RemiLehe:
While the `Particles` object takes the thermal velocities as argument, the `add_new_species` methods did not.

This PR fixes this. It adds the thermal velocities as arguments, and performs the corresponding boosted-frame transformations if needed (using an approximate formula). 

I also incorporated this change in on of the automated tests.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -640,7 +640,9 @@ class Simulation(object):
     def add_new_species( self, q, m, n=None, dens_func=None,
                             p_nz=None, p_nr=None, p_nt=None,
                             p_zmin=-np.inf, p_zmax=np.inf,
-                            p_rmin=0, p_rmax=np.inf, uz_m=0.,
+                            p_rmin=0, p_rmax=np.inf,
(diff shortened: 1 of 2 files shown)


MKirchen:
In order to avoid the approximation we would need to perform the Lorentz transformation when creating the particles in the continuous_injector, correct?

MKirchen:
I guess short term the solution you propose in the PR is sufficient. Long term, I think it would make more sense if the contious_injector itself handles the boost when creating the particles. (so the boost object is just passed to the Particles/injector with add_new_species not boosting quantities.) What do you think? (I would suggest merging the PR as it is now.)

RemiLehe:
Yeah, that makes, we could merge the current PR, and then think about a better way to do it in the future. Maybe we could pass the `BoundaryCommunicator` to `Particles` so that the positions (`p_zmin`, `p_zmax`) can be directly boosted inside the `Particles` object.

PR #278, opened 2018-06-26
enable checkpoints in particle-less simulations

hightower8083:


hightower8083:
currently saving checkpoint in simulations _without particles_ raises an error. Moreover, if such checkpoint is written and used to start a simulation _with particles_, it would also give an error. Such manipulations can be useful in simulations with externally defined laser fields, e.g. laser profiles which are measured at focus and require preparatory backward run.

This PR fixes this issue.

RemiLehe:
Thanks for this PR! I like the proposed changes.

@hightower8083 Although we discussed these changes off-line, I now wonder if we should not actually force the user to have the same number of species at the time of the restart (i.e. `0` species if ts.available_species is `None` and `len(ts.available_species)` otherwise), by raising an exception (e.g. `RunTimeError`). In this exception, we could let the user know that additional species can be added *after* the restart by using `add_new_species`. (in the case with no species, we could let the user know that removing species can be done with `sim.ptcl = []` before calling the restart function)

The main reason is that, if the user creates species *before* the restart, then the positions of the corresponding particles is based on the positions of the simulation box *before* the restart. Because the positions of the simulation box might change *after* the restart, this might mess the simulation, if these species are unaffected by the restart (as is the case in your code for `ts.available_species = None`)

RemiLehe:
@MKirchen @hightower8083 Would you be okay with the above, and with raising an Exception in the case where the number of species does not match?

hightower8083:
@RemiLehe , I guess you are right and this approach is indeed more safe. Also, you know, I prefer the manual species initialisation. I'll correct this behaviour.

RemiLehe:
Great! Thank you.

RemiLehe:
Yeah, that was a failure due to randomness: I recently added particle temperature in the continuous injection test, but this means that random fluctuation can cause the test to fail (because the tests checks that the density is uniform).  
I re-started the test and it is now passing.

RemiLehe:
Thanks for your contribution @hightower8083 !

PR #282, opened 2018-07-11
Time reverse

hightower8083:
Option to reverse laser/particle beam direction at the checkpoint save. Useful to propagate the beams with externally defined profiles, e.g. a laser measured at focal position can be rolled back (in vacuum) and then shot as being focused in a plasma.

### DIFF
--- a/fbpic/openpmd_diag/checkpoint_restart.py
+++ b/fbpic/openpmd_diag/checkpoint_restart.py
@@ -60,7 +60,7 @@ def set_periodic_checkpoint( sim, period ):
         sim.checkpoints.append(
             ParticleDiagnostic( period, particle_dict, write_dir=write_dir ) )
 
-def restart_from_checkpoint( sim, iteration=None ):
+def restart_from_checkpoint( sim, iteration=None, reverse_time=False ):
     """
     Fills the Simulation object `sim` with data saved in a checkpoint.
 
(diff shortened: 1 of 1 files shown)


RemiLehe:
Thanks for this PR! I'll try to have a look today.

MKirchen:
Is „time reverse“ the correct wording for this functionality - it sounds maybe a little bit too general? Also, I am not sure if it is maybe to specialized to be included in the checkpoint restart, but rather it might be a separate function/method (its nice to keep things simple and separated)? 

Moreover, for the use case discussed, maybe an option would be to just in general support performing simulations „backwards“ (step(-100)) and then forward again after adding the plasma (step(+100...)). However, making this work might require some efforts. What do you think @rlehe?

MKirchen:
Following an offline discussion with @RemiLehe, I am in favor of his proposed solution „sim.reverse_time()“. And by the way, thanks for working on this @hightower8083

RemiLehe:
@hightower8083 Would it be okay with you to implement a method `sim.reverse_time` instead? 
If yes, do you want to do a new branch and a new PR for that? (Since it seems that the code would be modified in a different way, in this case.)

hightower8083:
Thanx for such an interesting feedback guys, and thanx for being interested in this contribution.

Getting sim.reverse_time as a separate method is fine with me -- where do you think is best to put it? 
logically it would fit at something like `lpa_utils/generic.py` but its up to you. I'll open a new branch for this.

BTW, the idea of negative `dt` is kinda cool -- would be a really pro feature B-)

RemiLehe:
Yeah, I think this could be either directly in `main.py` as a method of the `Simulation` class (if we want to call it with `sim.reverse_time()`) or in `lpa_utils` as a simple function (if we want to call it with `reverse_time(sim)`).

I think that I have a slight preference for the first option (even though `main.py` is already quite long). @MKirchen Any opinion?

PR #283, opened 2018-07-12
[WIP] Allow one electron species per ionization level

RemiLehe:
Several users requested that the electrons coming from different ionization level be distinguishable.

In this PR, this is done by using a separate electron species for each ionization level. (The option to use a single electron species - the previous behavior - is still kept, since using separate electron species typically impact the performance.)

The current branch is working but I am not a fan of the API I came up with. It is very error-prone (i.e. the user has to setup the species in the right order, and it is easy to get confused about which species corresponds to which ionization level.) @MKirchen @soerenjalas @hightower8083 Any suggestion for a better API?

### DIFF
--- a/fbpic/particles/elementary_process/cuda_numba_utils.py
+++ b/fbpic/particles/elementary_process/cuda_numba_utils.py
@@ -31,7 +31,7 @@ def perform_cumsum( input_array ):
     element is 0 and its last element is the total sum of `input_array`)
     """
     cumulative_array = np.zeros( len(input_array)+1, dtype=np.int64 )
-    np.cumsum( input_array, out=cumulative_array[1:] )
+    np.cumsum( input_array, out=cumulative_array[1:], axis=-1 )
     return( cumulative_array )
 
 def reallocate_and_copy_old( species, use_cuda, old_Ntot, new_Ntot ):
@@ -56,7 +56,7 @@ def reallocate_and_copy_old( species, use_cuda, old_Ntot, new_Ntot ):
     """
     # Check if the data is on the GPU
     data_on_gpu = (type(species.w) is not np.ndarray)
-    
+
     # On GPU, use one thread per particle
     if data_on_gpu:
         ptcl_grid_1d, ptcl_block_1d = cuda_tpb_bpg_1d( old_Ntot )
--- a/fbpic/particles/elementary_process/cuda_numba_utils.py
+++ b/fbpic/particles/elementary_process/cuda_numba_utils.py
@@ -13,15 +13,18 @@ from fbpic.utils.cuda import cuda_installed
 if cuda_installed:
     from fbpic.utils.cuda import cuda, cuda_tpb_bpg_1d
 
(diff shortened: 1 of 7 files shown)


hightower8083:
seems nice, and API seems logical enough for me -- its an advanced feature and this API is maximally simple and complete, which is important.

Since this modif concerns the  performance, I think it would be a good occasion to grant users access to the `level_max` -- sometimes one does not need to consider levels above some achievable `a0`, and in this case it would limit number of unless species created. Keeping this argument optional `level_max=None`, would set it to maximal level. 

Another question would it help in terms of performance to use common species? One practical  example for the Nitrogen:
```
elec_L_shell = sim.add_new_species( q=-e, m=m_e )
elec_K_shell= sim.add_new_species( q=-e, m=m_e )
target_species = [elec_L_shell, elec_L_shell, elec_L_shell, 
                              elec_L_shell, elec_L_shell, elec_K_shell, 
                              elec_K_shell ]
```

BTW, why in the test you need this dictionary  `elec_species_dict`? It seems you could remove some complexity by using directly the `target_species` list, e.g.:
```
 target_species = []
        for i_level in range(level_start, 7):
            target_species.append(sim.add_new_species( q=-e, m=m_e ))
```
[...]
```
assert np.allclose( target_species[i_level-level_start].w.sum(), n_N )
```

RemiLehe:
Thanks, these are very valid points!

- Regarding `level_max`: this is a good point ; we could add this. Since this is somewhat orthogonal to the current PR, I would still do a separate PR for that.

- Regarding helping performance: yes, I think that it would help to use as few species as possible, since the drop in performance is mainly due to the fact that we have to call kernels (gathering, pushing, deposition) one per species. And passing the same species several times in the list should work.

- Regarding the dictionaries: intuitively this how I thought users would typically use it. The problem with lists is that it is very easy to forget to subtract `level_start` when doing 
```
target_species[i_level-level_start]
```
and then it is possible to not even notice this problem when looking at the results of the simulation...
Dictionaries are less error prone, since you can directly address them with `i_level` (even if the ionizable levels do not start at 0)
In fact, for this reason, I now think that it would maybe be better to modify the API to pass a *dictionary* as `target_species`. For instance, in your above example, (and starting ionizing at level 2), this would be:
```
target_species = {
      2: elec_L_shell,
      3: elec_L_shell,
      4: elec_L_shell,
      5: elec_K_shell,
      6: elec_K_shell }
ions.make_ionizable( 'N', level_start=2, target_species=target_species)
```
Would that be okay with you? If yes, then I'll update the PR accordingly.

hightower8083:
sure having `target_species` as a dictionary is fine(you know I'm a fan of `dicts`) -- it was just confusing to have mixed dict-list style for the same thing, and these tests are often good as examples for copy-pasting, so its better to be as clear as possible. Having `target_species ` as a dictionary is fine. 

I'll do so runs with this feature asap, and will update you 👍

RemiLehe:
Sounds good. Then I'll update the PR with the new `dict`-only API.

PR #284, opened 2018-07-15
Time reverse 2

hightower8083:
Method to reverse laser/particle beam direction. Useful to propagate the beams with externally defined profiles, e.g. a laser measured at focal position can be rolled back (in vacuum) and then shot as being focused in a plasma.


RemiLehe:
Thanks for this new PR!
I'll close #282 as a result.

PR #286, opened 2018-07-17
Handling deposition of the chi factor in the envelope model

PierrePelletier:
This PR implements the deposition of chi on the grids to take into account the effects of the particles on the envelope field, following the same model as the deposition of rho. It also adds new automated tests:
- test_envelope_group_velocity which checks that a linear laser pulse propagates at the correct velocity in a plasma for modes 0 and 1
- test_linear_wakefield_envelope which is the analog of test_linear_wakefield for the envelope model


RemiLehe:
Thanks for this pull request!
I have a few stylistic comment (see inline comments).

In addition, I fixed a few issues on GPU in this PR: https://github.com/PierrePelletier/fbpic/pull/1
If you agree with these comments, you can merge that PR, and then I'll merge the current PR into the official repository.

PR #287, opened 2018-07-18
Close #267 Per-species density diagnostic

### LINKED ISSUE #267: Per-species density diagnostics
Several users said that it would be useful to have a diagnostic that calculates the density of a given species on a grid (as opposed to having the overall charge density, i.e. the sum of all species).

This could be done by implementing a new diagnostic, e.g. 
```
SpeciesDensityDiagnostic( species, ... )
```
This would require to allocate dedicated arrays inside this object, and maybe reuse the existing deposition functions to efficient fill these arrays on the GPU/CPU. The arrays would then be stored in an openPMD file as a regular field.

The real difficulty is maybe if we wish to do this for back-transformed diagnostics (boosted frame to lab frame)...

RemiLehe:
This PR implements a new type of diagnostic: `ParticleDensityDiagnostic`. This outputs the density of one given species, and can be useful e.g. in the case where the plasma is neutral (so as to still see the plasma profile in the output).

Regarding the API: I included this diagnostics in the example script for ionization. I also added the diagnostics in the documentation. Please let me know if you have any suggestion regarding the API.

Regarding the implementation: in order to avoid having to add a new array, the density of the requested species is deposited directly to the array `rho` in interpolation space. However, this will not perturb the on-going simulation, because the deposition does not update the spectral space in this case. (And the simulation only cares about the spectral components `rho_next` and `rho_prev`). In practice, this was done by generalizing `sim.deposit` so as to be able to pass a list of species, and to indicate whether or not the spectral space should be updated.

The new `sim.deposit` is now also used for the space charge calculation, as it saves a few lines.

The implementation was, in the end, quite simple, but has a few limitations:
- This writes the *charge density* ; for neutral particles, it will not produce anything.
- The produced density is not smoothed (because we do not go through spectral space)

### DIFF
--- a/docs/source/example_input/ionization_script.py
+++ b/docs/source/example_input/ionization_script.py
@@ -26,8 +26,9 @@ from scipy.constants import c, e, m_e, m_p
 # Import the relevant structures from fbpic
 from fbpic.main import Simulation
 from fbpic.lpa_utils.laser import add_laser
-from fbpic.openpmd_diag import FieldDiagnostic, ParticleDiagnostic, \
-     set_periodic_checkpoint, restart_from_checkpoint
+from fbpic.openpmd_diag import FieldDiagnostic, \
+    ParticleDiagnostic, ParticleDensityDiagnostic, \
+    set_periodic_checkpoint, restart_from_checkpoint
 
 # ----------
 # Parameters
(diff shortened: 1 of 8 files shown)


fbpic/main.py
```diff
@@ -521,20 +522,29 @@ def deposit( self, fieldtype, exchange=False ):
             Whether to exchange guard cells via MPI before transforming
             the fields to the spectral grid. (The corresponding flag in
             fld.exchanged_source is set accordingly.)
+
+        TODO
```
MKirchen: I guess this should be deleted?

docs/source/example_input/ionization_script.py
```diff
@@ -147,12 +148,17 @@ def dens_func( z, r ) :
...
     sim.diags = [
                 FieldDiagnostic( diag_period, sim.fld, comm=sim.comm ),
                 ParticleDiagnostic( diag_period,
                     {"electrons from N": elec_from_N, "electrons": elec},
-                    comm=sim.comm )
+                    comm=sim.comm ),
+                # Since rho from `FieldDiagnostic` is 0 almost everywhere
+                # (neutral plasma), it is useful to see the charge density
+                # of individual particles
+                ParticleDensityDiagnostic( diag_period, sim,
```
MKirchen: Maybe we should call it ParticleChargeDensityDiagnostic. Although this name is very long, it should be very clear to the user that the output density is charge dependent.

RemiLehe:
Thanks for your comments! I have updated the PR accordingly.

PR #289, opened 2018-07-20
Fix space-charge fields for high-energy bunch with emittance

RemiLehe:
A user recently reported that he was seeing issues in the initial space-charge fields of a high-energy bunch (GeV scale) with non-zero emittance.

When investigating this, I found that this was caused by the use of the transverse vector potential in the space-charge field calculation (which is proportional to the amplitude of the transverse velocities in the bunch - hence the importance of the non-zero emittance). In most other PIC codes, the transverse vector potential is neglected (because the transverse velocities in the bunch are typically small).

While I am not sure I fully understand the specific reason for this bug, I would recommend that FBPIC also neglect the transverse vector potential (just like most PIC codes) - at least until the problem is better understood. This is what this PR does.

Here are the space-charge fields in the **current dev branch** (for a 2 GeV bunch with 5 mm.mrad):
png image 600 x 500 pixels

Here are the fields with the current PR:
png image 600 x 500 pixels


RemiLehe:
OK, thanks! I'll try to do this today.

PR #294, opened 2018-09-12
Calculate number of iterations in boosted frame

MKirchen:
This PR adds a function to automatically calculate the required number of timesteps int the boosted frame. Also, the documentation is updated to explain the calculation to the user.

ToDo:
- [x] Implement code changes
- [x] Discuss different definitions for the calculation

Comments:
- Added function to BoostConverter that calculates the estimated interaction time in the boosted frame
- Added documentation about speedup / interaction time calculation and resolution in the boosted frame
- Modified example scripts to automatically calculate the number of iterations based on the interaction time (e.g. time for the moving window to cross the plasma)
- Thereby, I realised that the outputs of the example files are not very user friendly. The standard lwfa example is too short to see a nice wake and the boosted frame example has not the best parameters to see a nice wake in the output. I modified the parameters of the examples slightly and made them run longer by default so that the user directly gets more useful outputs when running the examples for the first time.
- I modified the test_example_docs_script.py file and reduced the number of iterations during the test so that the tests still run fast enough.


docs/source/advanced/boosted_frame.rst
```diff
@@ -184,3 +197,136 @@ class to a velocity close to:
...
+occurs at a :math:`\gamma_b < \gamma_{wake}` before the simulation
+slows down again.
+
+.. note::
+
+    As highlighted in the above example, there is an optimum
+    :math:`\gamma_b` for which the interaction time is minimized for a
+    given simulation setup. The speed up of a boosted frame
+    simulation will reverse for too high values of :math:`\gamma_b`.
+    In practice, it is often advisable to choose :math:`\gamma_b < \gamma_{wake}/2`
```
MKirchen: Replace with new formulation: 

In practice, it is often advisable to fulfil :math:`\gamma_b^2 < L_{plasma}/l_{window}` for maximum efficiency and :math:`\gamma_b < \gamma_{wake}/2` to properly resolve the physical objects (see next section).

PR #296, opened 2018-09-27
Minor spelling

fractional-ray:
I think harware is supposed to be hardware.

thanks,
-isaiah


RemiLehe:
Good catch! Thanks for fixing it :)

PR #298, opened 2018-10-10
[WIP] Load balance simulation box with damp cells

RemiLehe:


RemiLehe:
This expected to give better performance, but did not in practice.

PR #299, opened 2018-10-15
Fix precision limitation for flattened gaussian

soerenjalas:
When using the flattened Gaussian profile with very high orders (~200) the class is returning nan as the E field. This is because the method runs into machine precision problems. This PR fixes the problem by slightly changing the order of calculations.


RemiLehe:
Great! Thanks for fixing this.

PR #300, opened 2018-10-17
Add injection area to damping region

MKirchen:
Fix a bug when the damping region is not entirely filled with plasma after exchange_period steps.

Before this PR, new particles where added only in the physical domain + ~ exchange_period cells ahead of the physical domain. Thus not the entire damping region was filled with plasma. In some cases, non-zero fields can appear in the damping region which can interact with particles injected at different times, causing weird oscillations in e.g. the charge density.

This PR changes this by adding new particles up to the physical domain + the damping region + an additional injection area with n_inject cells. A new parameter n_inject has been added to the Simulation object and the BoundaryCommunicator. It can be set manually but defaults to n_guard/2. It is considered to be part of the damping region.

With the new injection area, particles will never be injected in the guard region and it is ensured that newly added particles do not see any fields as they are injected only in the injection area.

In addition, the pure damping region has been changed to 50:50 (damping_function:no_damping) compared to 33:66 before.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -49,7 +49,7 @@ class Simulation(object):
                  n_order=-1, dens_func=None, filter_currents=True,
                  v_comoving=None, use_galilean=True,
                  initialize_ions=False, use_cuda=False,
-                 n_guard=None, n_damp=64, exchange_period=None,
+                 n_guard=None, n_damp=64, n_inject=None, exchange_period=None,
                  current_correction='curl-free', boundaries='periodic',
                  gamma_boost=None, use_all_mpi_ranks=True,
                  particle_shape='linear', verbose_level=1 ):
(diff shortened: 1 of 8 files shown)


fbpic/main.py
```diff
@@ -49,7 +49,7 @@ def __init__(self, Nz, zmax, Nr, rmax, Nm, dt,
                  n_order=-1, dens_func=None, filter_currents=True,
                  v_comoving=None, use_galilean=True,
                  initialize_ions=False, use_cuda=False,
-                 n_guard=None, n_damp=64, exchange_period=None,
+                 n_guard=None, n_damp=64, n_inject=None, exchange_period=None,
```
RemiLehe: I am wondering whether this parameter should be exposed in the API of the `Simulation` object. 
I think that it might confuse the users to have too many parameters (e.g. in the automatic documentation), and I think that there would very rarely be a case where we might want to change `n_inject`. 
@MKirchen What do you think about dropping this argument and always use the default number for `n_inject` (`n_guard/2`)

MKirchen: Hey, that was my initial plan, but then I decided to explicitly add it to the Simulation object so that the user always has full control of the total number of cells if he wants (or at least can deduce from the parameters of the Simulation object what the total or local number of cells will be). Also if someone decides to use a custom exchange_period, he might want to use a custom number of injection cells as well. However, I do not have a strong opinion on this. I can also remove the parameter from the Simulation object.

PR #301, opened 2018-10-18
[WIP] Fix injection bug by restricting field gathering in the right damping region

MKirchen:
This PR limits the gathering of the fields onto the Particles to the physical domain at the right boundary of the global simulation box. 

Now particles are only injected in the damping region at the right side of the box and it is additionally made sure that the damping region is large enough such that no particles get injected in the guard region.

By not gathering the fields onto the particles in the damping region, problems caused by injecting particles in batches at different iterations and positions in the damping region should be resolved.

(Still needs testing)


MKirchen:
This was redone differently in PR #300

PR #302, opened 2018-10-19
Correct Gouy phase of Laguerre-Gaussian modes

RemiLehe:
In the implementation of the Laguerre-Gaussian modes, the sign of the Gouy phase is incorrect. 

This was not detected previously, because the phase is not checked in the automated tests. 

This bug does affect the flattened Laguerre-Gauss profile, because the relative phase of the different Laguerre-Gauss modes that compose the pulse is then incorrect and produces the wrong overall profile. This was again not detected by the automated tests because they only initialized the flattened Gaussian at focus, where the Gouy phase is 0. Therefore, this pull request also modifies the flattened Gaussian test, so as to initialize the pulse away from focus. With the current `dev` branch, this test fails, but with the correction of this PR, it passes.


MKirchen:
Thanks for the fix!

soerenjalas:
Thanks a lot!

PR #303, opened 2018-10-20
[WIP] Prevent gathering of the fields outside of the physical box

RemiLehe:
As discussed offline with @MKirchen, particles should not gather fields from outside of the box (e.g. from the damp cells). This can especially influence the continuously-injected particles, before they enter the box.

This PR prevents the particles from gathering outside of the physical box.


RemiLehe:
The problem that this PR tries to fix has been fixed by PR #300

PR #305, opened 2018-10-26
Close #304: Allow changing checkpoint saving folder

### LINKED ISSUE #304: Allow changing checkpoint saving folder
Hey, 

Currently, it is possible to change the output folder of normal diagnostics, which is incredibly useful to perform parameter scans within one run script. However, it would be good if one could also set the checkpoint folder. Currently, if one tries to run another simulation in the same runscript with a different resolution, an error about sizes not matching will be given once a checkpoint is about to be written.

I assume this would involve not massive changes in the checkpoint_restart.py file?

Cheers,
Kris

RemiLehe:
This PR adds an optional argument for the functions that write and read the checkpoints. The optional argument points to the path where the checkpoints should be stored.

I also modified the automated tests in order to have them test that new functionality.


MKirchen:
Great as always 👍

PR #309, opened 2018-11-09
Fix parallel laser deposition

RemiLehe:
When the laser antenna crosses the boundary between 2 MPI procs, the current correction gets crazy. This is because rho^{n} is deposited on one MPI rank and rho^{n+1} is deposited on the other MPI rank.

In order to avoid this problem, this PR adds a flag that controls whether the antenna deposits on the current MPI proc (reminder: all MPI rank have a copy of the virtual particles of the antenna ; unlike the actual physical macroparticles).
This flag can only be changed at the beginning of a PIC iteration (i.e. at the same time as we exchange the physical particles between MPI ranks). That way, rho^n and rho^{n+1} are always deposited on the same MPI rank.

For instance, here is the Ez field of an injected laser, around the time when the antenna crosses the boundary between the 2 MPI ranks (red line) ; the current correction is the `cross-deposition`, which should be local ;  the laser is only emitted on the right-hand side because this is a boosted-frame simulation and the antenna is moving close to the speed of light (to the left). The upper plot is the current `dev` branch (unphysical fields in the left hand side) while the bottom plot is this PR:
deposition

### DIFF
--- a/fbpic/lpa_utils/laser/antenna_injection.py
+++ b/fbpic/lpa_utils/laser/antenna_injection.py
@@ -156,6 +156,10 @@ class LaserAntenna( object ):
         elif v_antenna != 0:
             self.vz += v_antenna
 
+        # Register whether the antenna deposits on the local domain
+        # (gets updated by `update_current_rank`)
+        self.deposit_on_this_rank = False
+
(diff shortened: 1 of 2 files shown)


MKirchen:
Thanks for fixing this!

PR #310, opened 2018-11-19
Fix continuous injection of plasma when no plasma is originally in the box

RemiLehe:
PR #300 introduced a small bug in the continuous injection routines: when there is no plasma in the box at t=0, the new plasma will only start being injected *after* the damp cells and injection cells, instead of being injected immediately after the end of the physical domain. As a result, simulations typically feature an absence of plasma in the first microns of the simulation.

This PR fixes the bug.
In addition, it adds an automated test to catch similar errors in the future (up to now, the automated tests only checked cases where there is some plasma at t=0 in the box). I also refactored a bit the continous injection test, in order to avoid code duplication.

### DIFF
--- a/fbpic/particles/injection/continuous_injection.py
+++ b/fbpic/particles/injection/continuous_injection.py
@@ -74,7 +74,7 @@ class ContinuousInjector( object ):
         dt: float (in s)
             Timestep of the simulation
         """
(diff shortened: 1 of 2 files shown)


MKirchen:
Thanks for finding and fixing this!

PR #311, opened 2018-11-20
Fix Travis CI tests by limiting the number of threads running simultaneously.

RemiLehe:
It seems that something recently changed on Travis CI, and is making the simulations *very slow* when using more than 2 threads simultaneously.

This impacted the tests that use MPI, since we were running 2 MPI with 2 threads each (4 threads total running simultaneously). These tests were taking >5x longer, resulting in a timeout for the Travis build.

This PR fixes the issue by imposing a single thread when running the MPI tests.


MKirchen:
Thanks for fixing this!

PR #314, opened 2018-12-06
Subsampling

omriseemann:
Added "subsampling_fraction" to apply_selection function.
select particles randomly based on on subsampling_fraction.
numpy module.
checked with pep8, pyflakes and tested with setup.py test

### DIFF
--- a/fbpic/openpmd_diag/particle_diag.py
+++ b/fbpic/openpmd_diag/particle_diag.py
@@ -18,7 +18,8 @@ class ParticleDiagnostic(OpenPMDDiagnostic) :
 
     def __init__(self, period, species = {"electrons": None}, comm=None,
         particle_data=["position", "momentum", "weighting"],
-        select=None, write_dir=None, iteration_min=0, iteration_max=np.inf ) :
+        select=None, write_dir=None, iteration_min=0, iteration_max=np.inf, 
+        subsampling_fraction=None ) :
         """
         Initialize the particle diagnostics.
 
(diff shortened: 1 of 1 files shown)


RemiLehe:
Thanks a lot for the PR!

Regarding the API: I see that the user should pass the `subsampling_fraction` in the `select` dictionary. While I agree that this minimizes changes to the code, I also think it would result in a somewhat less consistent API (in the sense that sub-sampling is a somewhat different operation than rule-based selection). 

For this reason, I think I would prefer it if we introduce `subsampling_fraction` as an entirely new argument to `ParticleDiagnostic` (which would be `None` by default). `subsampling_fraction` would then be registered inside the `__init__` function as `self.subsampling_fraction` and used inside `apply_selection` (as a new argument as well). 
In this case, the code that you added would not be under the `if self.select is not None` but inside a new clause `if self.subsampling_fraction is not None`. 

Would you agree with the above changes? If yes, could you modify the PR accordingly?

omriseemann:
yes.

omriseemann:
Hi,
I finished the commit, Changed the code according to the suggestion:
Put new parameter in init function of ParticleDiagnostic
and modified apply_selection accordingly.

test2
test

RemiLehe:
Great, thanks a lot for these changes, and sorry for the late reply (I have been traveling for the last few days).

@MKirchen Do you also agree with this PR? If yes, we can merge.

MKirchen:
Thanks for the PR!

PR #315, opened 2018-12-20
Added more publications to README

MKirchen:
As discussed offline with @RemiLehe, I added additional publications with links to the Readme such that new users can more easily find research papers related to the algorithms of FBPIC. Additionally, we now ask the user to also cite the additional references in their publications if they used the more advanced features of FBPIC.


MKirchen:
Sorry, totally forgot about the index.rst!

I added the changes. Ready to review again. Thanks!

PR #316, opened 2019-01-07
Added writing and selection based on particle gamma to the particle diagnostic

syoffe:
Hello,

I have added saving particle (Lorentz) gamma to the particle diagnostic, as well as particle selection based on gamma. When momentum is requested (`'momentum' in particle_data = True`) then the gamma is also saved in a similar manner to the particle weights. Gamma is obtained from the `inv_gamma` variable stored by each particle.

This is my first pull request, so I apologise in advance if I have not done this correctly!

Best wishes,
Sam


MKirchen:
Hey Sam, thanks for the PR! Could you please check why the tests are failing? (I've set your title to [WIP] (work in progress) so that we are not confused and merge too early by mistake)

.gitignore
```diff
@@ -10,3 +10,5 @@ dist/
 fbpic.egg-info/
 .cache/
 .eggs/
+
+.idea/
```
RemiLehe: Thanks for the PR! Could you revert the above changes to the `.gitignore`? Thanks!

fbpic/openpmd_diag/particle_diag.py
```diff
@@ -92,7 +92,7 @@ def __init__(self, period, species = {"electrons": None}, comm=None,
                 if quantity == "position":
                     self.array_quantities_dict[species_name] += ['x','y','z']
                 elif quantity == "momentum":
-                    self.array_quantities_dict[species_name] += ['ux','uy','uz']
+                    self.array_quantities_dict[species_name] += ['ux','uy','uz','gamma']
```
RemiLehe: Hm, I think that this changes the default behavior, which is not ideal...
Would you mind, adding an extra `elif quantity == "gamma":`, so that `gamma` is not written when the user enters `"momentum"`, but is only written if the user adds `"gamma"` to the `particle_data` list? (and could you modify the docstring for `particle_data` accordingly`?)

syoffe: Yeah sure, actually the code doesn't need an extra `elif` since the final `else` would already append it.

.gitignore
syoffe: Will do.

syoffe:
I've updated the PR based on Remi's comments and it should now pass the tests (at least it passed all tests locally).

syoffe:
I'm a bit confused by the failed tests - all tests passed fine on my local machine (top image, no GPU) and seemed to be passing fine on Travis CI before it was aborted (lower image).

tests

travisci

RemiLehe:
The tests seems to fail on Travis even for the `dev` branch: 
https://travis-ci.org/RemiLehe/fbpic/jobs/476624607
This is strange ; I'll try to investigate.

RemiLehe:
So it seems that Numba 0.42 causes the current deposition with cubic shape factor to crash on CPU (the charge deposition is fine ; current deposition with linear shape factor is fine ; deposition on GPU is fine).

I expect that this might take some time to solve.

@MKirchen @soerenjalas In the meanwhile, I'll do a PR similar to https://github.com/fbpic/fbpic/pull/123/files so as to impose numba 0.41. I suggest that we also do a quick bug-fix release so that users don't get surprised by this crash with numba 0.42. 

@syoffe I think that this PR will have to wait until the bug-fix release (shouldn't take more than a few days)

syoffe:
@RemiLehe No problem at all, I must admit to being slightly relieved that it's not just my poor coding!

syoffe:
Yay, all checks have passed :-)

PR #319, opened 2019-01-17
Allow FBPIC to work with numba 0.42

RemiLehe:
It is seems that issue with `numba 0.42` (see #317) was related to the fact the functions for the shape factors (in the deposition kernels), contained a return statement within an `if` condition. (see the functions `Sz_linear`, etc. in `fbpic/particles/deposition/threading_methods.py`)

In fact, these functions do seem a bit weird, since they have cases that result in no `return` statement (hence what value do they return in this case?!) : see for instance `Sz_linear` for the case `index=2` - although it is never used in practice.

I modified these functions, so as to take the `return` statement outside of the `if` condition (see corresponding code below). 
Although the GPU functions were not causing any issue with numba 0.42, I modified them accordingly, for consistency. I tested the performance on a GTX 1080 Ti and did not see any significant change.

Since the CPU functions now work properly with numba 0.42, I removed all the restrictions imposed by #317 .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,7 @@ before_install:
 
 # Install packages
 install:
-  - conda install --yes numba==0.41 scipy h5py mkl
+  - conda install --yes numba scipy h5py mkl
   - conda install --yes -c conda-forge mpi4py mpich
   - pip install pyflakes
   - python setup.py install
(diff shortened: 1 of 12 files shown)


MKirchen:
Okay with the changes! Looks better now. Probably the direct returns where implemented by Kevin to avoid any more registers in the GPU kernels. But I guess nothing is stored in registers anyways as those functions are inlined.

PR #321, opened 2019-01-29
[WIP] On-axis charge and current deposition

MKirchen:
This PR changes the deposition of charge and current on-axis to match the definition of the various fields on axis.

In the current dev branch, we use an alternative definition that was implemented to ensure correct physical results. For every charge or current to be deposited close to the axis (below the first cell index at dr/2), the contribution was always subtracted from the first cell (value*-1 is added). This way, small transverse displacements of the charge of a homogeneous plasma would not lead to a non-homogeneous charge density on-axis.

In this PR, the charge and current deposited close to the axis is either added or subtracted from the first cell depending on the specific component and mode of the field. The on-axis deposition would then be the same as the on-axis gathering and match the correct definition of the fields on-axis.
This PR might improve the on-axis accuracy of simulation results for some cases, but it is work in progress and still needs testing.

Edit: Added possibility to do a Verboncoeur-type correction

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -50,7 +50,8 @@ class Fields(object) :
     def __init__( self, Nz, zmax, Nr, rmax, Nm, dt, zmin=0.,
                   n_order=-1, v_comoving=None, use_galilean=True,
                   current_correction='cross-deposition', use_cuda=False,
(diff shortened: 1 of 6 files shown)


MKirchen:
This PR will be reopened in with an updated implementation

PR #322, opened 2019-01-30
Add installation instructions for Summit

RemiLehe:
Here are the instructions for running on Summit (PowerPC with V100 GPUs).


MKirchen:
Nice, thanks for those instructions!

PR #325, opened 2019-02-01
Close #323: Avoid negative gamma when initialising gaussian bunch

### LINKED ISSUE #323: `add_elec_bunch_gaussian` can generate particles with negative `gamma`
Because of the way in which `add_elec_bunch_gaussian` generates `gamma`, some particles can have negative `gamma` (unphysical)!

One solution would be to generate `uz` with `np.random` (instead of `gamma`), and then calculate the corresponding `gamma`.
Another solution would be use `gamma = np.max(1, np.random...`.

In both cases however, the standard deviation of the distribution of `gamma` would not match the `sigma_gamma` requested by the user. (Although they would only significantly differ in the case of very large energy spread.)

@MKirchen @soerenjalas Any preference, or any suggestion on how to do it differently.

soerenjalas:
This is a suggestion to solve the issue brought up by @RemiLehe.
The approach would be to remove those particles that result in unphysical gammas. This would of cause change the mean gamma and the std of the distribution but I see no way of avoiding that. 

It is not straightforward to only remove the particles that have a gamma < 1, since there can be the case where ux and uy which are generated according to the emittance of the beam result in a gamma that is larger than the randomly generated gamma and thous would require an imaginary uz for uz,ux,uy and gamma to be consistent.
Therefore, similar as before ux, uy and gamma are calculated independently and from that the square of uz is calculated. Then all the particles are dropped which have a uz**2 < 1. If that happens the user is issued a warning stating how many particles were removed.

The below shows what happens to the spectrum when it gets truncated. The spectra all have the same  sigma_gamma and different gamma0. There is basically a smooth transition so that dQ/d_gamma is 0 at gamma = 1.  
Right now the implementation is such that the charge is kept constant when removing the particles and therefore the spectral density increases when the beam is truncated.

trunc_bunch

### DIFF
--- a/fbpic/lpa_utils/bunch.py
+++ b/fbpic/lpa_utils/bunch.py
@@ -94,6 +94,7 @@ def add_elec_bunch( sim, gamma0, n_e, p_zmin, p_zmax, p_rmin, p_rmax,
     # Get the corresponding space-charge fields
     get_space_charge_fields( sim, relat_elec, direction=direction )
 
+
 def add_elec_bunch_gaussian( sim, sig_r, sig_z, n_emit, gamma0,
                         sig_gamma, Q, N, tf=0., zf=0., boost=None,
                         save_beam=None, z_injection_plane=None ):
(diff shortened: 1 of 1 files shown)


fbpic/lpa_utils/bunch.py
```diff
@@ -153,38 +154,59 @@ def add_elec_bunch_gaussian( sim, sig_r, sig_z, n_emit, gamma0,
...
-    uz = np.sqrt((gamma**2-1) - ux**2 - uy**2)
-    # Get inverse gamma
-    inv_gamma = 1./gamma
+    uz_sqr = (gamma ** 2 - 1) - ux ** 2 - uy ** 2
+
+    # Check for unphysical particles with uz**2 < 0
+    mask = uz_sqr <= 0
+    N_remove = np.count_nonzero(mask)
+    if N_remove != 0:
+        print("Warning: Particles with uz**2<0 detected."
```
RemiLehe: Actually: could you use `warnings.warn(` instead of `print`? (see for instance https://github.com/fbpic/fbpic/blob/dev/fbpic/main.py#L203)
I think this is now our preferred way of sending messages to the user.
Note: this requires `import warnings` at the beginning of the script.

fbpic/lpa_utils/bunch.py
```diff
@@ -153,38 +154,59 @@ def add_elec_bunch_gaussian( sim, sig_r, sig_z, n_emit, gamma0,
...
-    # Now we imprint an energy spread on the gammas of each particle
+    # Generate Gaussian gamma distribution of the beam
     if sig_gamma > 0.:
         gamma = np.random.normal(gamma0, sig_gamma, N)
     else:
-        # Or set it to zero
+        # Zero energy spread beam
         gamma = np.full(N, gamma0)
         if sig_gamma < 0.:
             print("Warning: Negative energy spread sig_gamma detected."
```
RemiLehe: I know this was not modified by your PR, but could you use `warnings.warn` here instead of print?
(see other comment below)

fbpic/lpa_utils/bunch.py
```diff
@@ -153,38 +154,59 @@ def add_elec_bunch_gaussian( sim, sig_r, sig_z, n_emit, gamma0,
...
+
     # Finally we calculate the uz of each particle
     # from the gamma and the transverse momenta ux, uy
-    uz = np.sqrt((gamma**2-1) - ux**2 - uy**2)
-    # Get inverse gamma
-    inv_gamma = 1./gamma
+    uz_sqr = (gamma ** 2 - 1) - ux ** 2 - uy ** 2
+
+    # Check for unphysical particles with uz**2 < 0
+    mask = uz_sqr <= 0
```
RemiLehe: Actually, for readability, could you do `mask = uz_sqr >= 0`, and then `N_new = np.count_nonzero(mask)`, and `if N_new < N:`, and then `x = x[mask]`, etc.

I'm a bit picky here, but I find more intuitive :)

soerenjalas: I agree, the ~ operator is a little bit unreadable. I inverted the selection.

PR #327, opened 2019-02-02
option to define a 3D plasma density profile

hightower8083:
In some cases, user could benefit using multi-mode (`Nm`>2) simulations with asymmetries in the transverse plasma profile, e.g. some non-ideal guiding potential for the laser, however, presently `dens_func` can only be purely axisymmetric.

This PR adds an option to define `dens_func` with three arguments `x`, `y` and `z`, instead of two `z` and `r`. Number of arguments is implicitly retrieved by `inspect.signature` method, so that this modification is fully backward-compatible: `2 args => (z,r)` and `3 args => (x,y,z)`

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -704,9 +704,11 @@ class Simulation(object):
            If `n` is not None, evenly-spaced macroparticles will be generated.
 
         dens_func : callable, optional
-           A function of the form :
+           A function of two of three arguments of the form :
            def dens_func( z, r ) ...
-           where z and r are 1d arrays, and which returns
+           or
+           def dens_func( x, y, z ) ...
+           where x, y, z and r are 1d arrays, and which returns
            a 1d array containing the density *relative to n*
            (i.e. a number between 0 and 1) at the given positions
 
--- a/fbpic/particles/injection/continuous_injection.py
+++ b/fbpic/particles/injection/continuous_injection.py
@@ -8,6 +8,7 @@ It defines a class for continuous particle injection with a moving window.
 import warnings
 import numpy as np
 from scipy.constants import c
+from inspect import signature
 
 class ContinuousInjector( object ):
     """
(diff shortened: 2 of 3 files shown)


RemiLehe:
Thanks for this PR (sorry for taking so long to review it)!

After thinking about this, I would like to suggest a different approach: allow the user to pass either a *class* (that represents the plasma profile) or a function (current behavior, kept for backward compatibility). 
If the user passes a function, it has to be of the form `dens_func(z,r)` as before.
If the user passes a class, there is much more freedom (similar to passing a `laser_profile` class to [add_laser_pulse](https://fbpic.github.io/api_reference/lpa_utilities/laser.html#fbpic.lpa_utils.laser.add_laser_pulse)). The potential plasma-profile classes could be:
- from a 3D cartesian function (your case)
- from a [numexpr](https://numexpr.readthedocs.io/en/latest/user_guide.html) expression (nice because of automatic optimization/threading)
- from a file
etc...

The advantages compared to the current PR would be that it is:
- Closer to the PICMI standard (standard input format, that eventually would be the input of Warp, WarpX, FBPIC and others)
- Potentially more robust: for instance with the current PR, if the user passes the following function, it could cause problems:
```
def dens_func(z, r, plot=False):
      n = np.ones_like(z)
      if plot:
            plt.plot(z, n)
      return n
```
with `inspect`, FBPIC would think that the above function is to be treated as a function of `x`, `y`, `z`.

@hightower8083 If you agree with the above suggestion, I can close this PR and draft a new one with the above proposed changes.

hightower8083:
Yes this is a good idea to allow a more advanced interface compatible with WARPs. 

Anyway in our study, we do not use a 3D expression, but construct it from the the input file interpolated in the transverse plane, e.g.
```
fu_z = lambda z : np.interp(z, [0, entrance_ramp], [0,1], left=0, right=1)
n_cfd, x_cfd, y_cfd = np.load('pic_input_2d.npy')
fu_xy = RectBivariateSpline( y_cfd, x_cfd, n_cfd )

def dens_func(x, y, z ):
    val = np.empty_like(z)
    for ip in range(z.size):
        val[ip] =  fu_xy(y[ip], x[ip]) * fu_z(z[ip])
    return val
```
but having this in the form of a _class_ is a more clear and flexible way to work.

PS : Just for the note, the mentioned problem with incorrect args, can be easily handled, since `inspector` can also check the defined names of vars, so we could assure, that they are in the format of `x`, `y` and `z`

PR #330, opened 2019-02-12
Option for writing the fields at particle positions

danielseipt:
Introduces option to write the electric and magnetic fields at the particle positions.
The `particle_data` keyword of the `ParticleDiagnostic` class accepts the new value
`"fields"` (which is not given by default).
When `"fields"` is given, all cartesian components of the electric and magnetic field are
written for each particle under `fields/Ex`, `fields/Ey`etc


RemiLehe:
Thanks for this PR @danielseipt !
I implemented a few modifications and made a corresponding PR from my fork to your fork: 
https://github.com/danielseipt/fbpic/pull/1

Could you check that this is okay with you, and that the modification do not significant affect (or preclude) your use case?
If everything is okay, you can merge the above-mentioned PR into your fork. This will then automatically update the current PR, which I will then merge into the official `dev` branch of FBPIC.

danielseipt:
Thanks for cleaning up my code and making it openPMD-compliant @RemiLehe.

I made another small change to make the user experience maybe a bit more consistent:
Instead of the previous keyword `particle_data=["fields"]` I changed it to `particle_data=["E","B"]`.
This way it resembles the case of `"position"`, `"momentum"`, etc where the data can be accessed from the output files with the same keywords that are used to write them.

RemiLehe:
Great! Thanks again for this PR!

PR #334, opened 2019-03-07
problem in example for boosted frame LWFA

hightower8083:
Example script for the boosted frame LWFA simulations contains a bug with the lab-frame diagnostics, and some ambiguity in the comments. 
- bug: `v_window` is overwritten by conversion for the moving window (which could probably be better done inside `sim.set_moving_window` to avoid any doubts) and  lab-frame diagnostics receive `c` instead of the lab-frame `v_window` value. As a result, the lab-frame fields are cut at the front, and the diagnostics frame propagates 4.2 mm instead of 4.5 and.
- ambiguity: `diag_period` defines both -- diags in boosted frame and record period for the lab-frame ones. As a newbie in boosted-frame sims, I can say its a bit perturbing, so its better to use different names.

This PR addresses these issues. 

P.S. I've also taken liberty to set a formula for `dt_snapshot_lab` to match the simulation duration.


RemiLehe:
Thanks! This is very useful.

Quick question: could you post an image (e.g. screen shot) to show want you mean by "As a result, the lab-frame fields are cut at the front" (e.g. could you send the images for the cases where you pass `c` and `v_window` respectively)

hightower8083:
Sure, 

Here is a final snap from the current script

image

and here is the one with this PR

image

hightower8083:
The cutting here is progressive -- in the longer simulations, the unfilled region at the front gets bigger.

MKirchen:
Hey Igor, I agree with you that the current API is not ideal. In principle, I agree with your solution, however I would go even further and try to simplify the input script even more. For the user, it might probably be ideal to define the diagnostics always in terms of the interaction length and the total number of outputs independent of gamma_boost. The input script should be defined in a way such that the user always gets the same outputs at the same times and positions in the standard lab frame and in the lab frame using the boosted 

Also, I agree that the naming of the variables (e.g. Ntot_snapshots_lab) is not too nice (sorry for that).

Going forward with this PR, I would agree with merging the suggested changes, but I would favor if you would add even more changes if you have time. @rlehe What do you think?

RemiLehe:
@MKirchen Yes, I agree (provided that this is okay with @hightower8083 and he has the time): please suggest further modifications using inline comments for this PR.

hightower8083:
@MKirchen sure, i can revise it further -- can you explain more of what you have in mind?

docs/source/example_input/boosted_frame_script.py
```diff
@@ -137,19 +137,26 @@ def dens_func( z, r ):
...
-dt_snapshot_lab = 10*(zmax-zmin)/c
-track_bunch = False  # Whether to tag and track the particles of the bunch
-
 # The interaction length of the simulation (meters)
 L_interact = (p_zmax-p_zmin) # the plasma length
 # Interaction time (seconds) (to calculate number of PIC iterations)
 T_interact = boost.interaction_time( L_interact, (zmax-zmin), v_window )
 # (i.e. the time it takes for the moving window to slide across the plasma)
 
+## The diagnostics
```
MKirchen: Here, I could imagine something like this (I did not test the code - this is open for discussion):

```
# Number of diagnostics to output (including first output at t=0)
N_diag = 10+1
# Time interval between diagnostics (first at t=0, last at t=T_interact)
dt_diag = (L_interact + (zmax-zmin)) / v_window / (N_diag - 1)
# Period of the (boosted frame) diagnostics in timesteps
diag_period = T_interact / (N_diag - 1)
# Period of writing the cached, backtransformed lab frame diagnostics to disk 
write_period = 50
```

This would ensure that in the case of gamma_boost=1, the actual standard diagnostics are defined at the same times/positions as the lab frame output of any boosted frame simulation. (Makes it easy to compare results between lab and boosted frame runs).

The boosted frame diagnostics would always be equally distributed within the boosted interaction time.

If we agree to this (or something similar), I would suggest to change the other two example scripts accordingly.

docs/source/example_input/boosted_frame_script.py
```diff
@@ -175,21 +182,21 @@ def dens_func( z, r ):
...
                     {"electrons":sim.ptcl[0], "bunch":sim.ptcl[2]}, sim.comm),
-                  BoostedFieldDiagnostic( zmin, zmax, c, dt_snapshot_lab,
+                  BoostedFieldDiagnostic( zmin, zmax, v_window, dt_snapshot_lab,
                     Ntot_snapshot_lab, boost.gamma0, fieldtypes=['rho','E','B'],
-                    period=diag_period, fldobject=sim.fld, comm=sim.comm),
-                  BoostedParticleDiagnostic( zmin, zmax, c, dt_snapshot_lab,
-                    Ntot_snapshot_lab, boost.gamma0, diag_period, sim.fld,
-                    select={'uz':[0.,None]}, species={'electrons':sim.ptcl[2]},
-                    comm=sim.comm )
+                    period=record_period, fldobject=sim.fld, comm=sim.comm),
```
MKirchen: Here, I would go maybe with `write_period` instead of `record_period` as it more explicitly tells the user that something is written to disk. However, I am open to any other suggestions.

docs/source/example_input/boosted_frame_script.py
```diff
@@ -175,21 +182,21 @@ def dens_func( z, r ):
...
+                  BoostedParticleDiagnostic( zmin, zmax, v_window,
+                    dt_snapshot_lab, Ntot_snapshot_lab, boost.gamma0,
+                    record_period, sim.fld, select={'uz':[0.,None]},
+                    species={'electrons':sim.ptcl[2]}, comm=sim.comm )
                     ]
 
     # Number of iterations to perform
```
MKirchen: Given the suggested changes above, I think we need to add a +1 to the iterations to guarantee that the last output is always written, i.e.:

```
# Number of iterations to perform
N_step = int(T_interact/sim.dt)+1
```

docs/source/example_input/boosted_frame_script.py
MKirchen: Thinking further about this, `record_period` is maybe also okay as in the OpenPMD standard, written outputs are also called 'records'.

MKirchen:
> The cutting here is progressive -- in the longer simulations, the unfilled region at the front gets bigger.

I guess this is not really a 'bug' (?), but just a consequence of the slightly wrong definition of the velocity of the outputs in the input script - or am I wrong here?

docs/source/example_input/boosted_frame_script.py
hightower8083: done

docs/source/example_input/boosted_frame_script.py
hightower8083: I think `write_period` is fine and its clearer how it works this way

hightower8083:
I agree that bug rather signifies source related issues.

docs/source/example_input/boosted_frame_script.py
```diff
@@ -137,19 +137,26 @@ def dens_func( z, r ):
...
 # (i.e. the time it takes for the moving window to slide across the plasma)
 
+## The diagnostics
+
+# Number of diagnostics to output (including first output at t=0)
+N_diag = 10+1
+# Time interval between diagnostics (first at t=0, last at t=T_interact)
+dt_diag = (L_interact + (zmax-zmin)) / v_window / (N_diag - 1)
+# Period of the (boosted frame) diagnostics in timesteps
+diag_period = T_interact / (N_diag - 1)
```
RemiLehe: Hm, here `diag_period` should be `int( T_interact / ((N_diag - 1) * dt_boost) )`, but `dt_boost` is not very easy to obtain (we get it below, by using `sim.dt`, but this is not defined yet here) ; alternatively, we could add a method `boost.time` to the `BoostConverter`. @MKirchen @hightower8083 Any opinion?

RemiLehe:
I pushed a small fix for the automated tests. 
But the tests are still failing because of the definition of `diag_period` (see inline comment)

docs/source/example_input/boosted_frame_script.py
hightower8083: oups, haven’t noticed absence of dt in diag_period definition.. 
i think that adding an explicit time coverter for dt_boost calculation is a good idea.

MKirchen: Sorry, my fault. I agree that we should add a boost.time() method.

hightower8083:
here i've added the `dt_boost` conversion via `boost.copropag_length`.

MKirchen:
Thanks, I like your solution! If @RemiLehe agrees with it, we should also update the calculation of `N_step` and `diag_period` in the other example scripts, and then we can merge.

RemiLehe:
OK, I just merged the changes from #335 in order to fix the tests.

From off-line discussion with @MKirchen : I think we would prefer the option of specifying directly the `dt_snapshot` to the Diagnostics (as an alternative to specifying the `period`). I'll do a separate PR ; once this is merged, I'll update the current PR. @hightower8083 Would that be okay with you?

hightower8083:
Sure, it will be nice to have similar syntax everywhere. Anyway, my initial point was to make this example more clear, since we've used it to understand how things are to be done in the boosted simulations.

RemiLehe:
Hey, I did a few updates. @hightower8083 @MKirchen Could you check that this is okay with you?

Also, I think we should probably rename the `BoostedFrameFieldDiagnostic` as `BackTransformedFieldDiagnostic` ; otherwise, I think it might be confusing for a user to know whether the corresponding data is in the lab frame or boosted frame. But I can do this in a separate PR.
(In this case, we would of course keep the `BoostedFrameFieldDiagnostic` for backward compatibility ; but we would not emphasize it in the examples/documentation.)

hightower8083:
Looks fine to me! 
Don't have preferences for methods naming nor for the backward compatibility. I guess its always possible to add alias `BoostedFrameFieldDiagnostic=BackTransformedFieldDiagnostic` and keep them till `1.0`

PR #339, opened 2019-03-21
Rename `BoostedDiagnostic` as `BackTransformedDiagnostic`

RemiLehe:
I think that the names `BoostedFieldDiagnostic` and `BoostedParticleDiagnostic` might be confusing for some users, e.g. since they are dumping data in the *lab* frame, not the *boosted* frame.

Also, this odd name leads to confusing initializations, as in this example script:
https://github.com/fbpic/fbpic/blob/dev/docs/source/example_input/boosted_frame_script.py#L196-209
where `dt_boosted_...` is passed to the regular diagnostics, but `dt_lab_...` is passed to the `Boosted` diagnostics.

This PR solves this issue by renaming these classes. The `Boosted` classes are kept for backward compatibility, but are not explicitly documented anymore. All the tests and examples now use `BackTransformed`.


MKirchen:
Thanks for this! 👍

PR #341, opened 2019-03-22
removed unused `uz_m` parameter

hightower8083:
sorry for being so pedantic, but as `uz_m` is unused it may perturb someone who learns the code, moreover since it's only accepted by `add_new_species` which is not present.

Another thing, since users may use this script as a template for their own simulations, may be theta  ppc number could be also given in a generic form (with some  comment), e.g. `p_nt = 4*(Nm-1)+(Nm==1)` ?  This way they would avoid some nasty instabilities which are not always easy to see on the final result


RemiLehe:
Thanks for catching that!

Regarding the number of modes: this is tricky since it is not always clear how many modes we need. For instance, we noticed that in the boosted frame, we often need 6 particles in theta, even with `Nm=2`, while the above formula gives `4`. I admit I do not have a good understanding of the required number of particles in theta...

PR #343, opened 2019-03-23
Automatically update the documentation, when we update master

RemiLehe:
This PR sets up Travis CI, so that it will automatically update the [online documentation](https://fbpic.github.io/) whenever we update the `master` branch. (i.e. whenever we do a bew release)

This essentially follows [these instructions](https://gist.github.com/domenic/ec8b0fc8ab45f39403dd).

The main point is that Travis CI needs to have the permissions to `git push` to the [documentation repository (fbpic.github.io)](https://github.com/fbpic/fbpic.github.io). In order to do so, I created an RSA key pair:
- I added the public key as a deploy key in the settings of the repository `fbpic.github.io`. This means that this repository will accept `git push` if it uses SSH with the private key.
- The private key is stored on the `fbpic` repository (as part of this PR: `deploy_key.enc`), but is encrypted. My Travis CI account maintains [secret environment variables](https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings) that allow it to decrypt this key (I added them through the Travis command line interface - they show up in the settings of `fbpic/fbpic` on my Travis account). These secret environment variables are only accessible when a Travis CI build is triggered by the official repository (`fbpic/fbpic`), not by forks (e.g. `RemiLehe/fbpic`). Thus, even though a Travis build on a Fork may try to update the documentation, it will not be able to decrypt the `deploy_key.enc`, and thus it will not be able to do `git push`.


deploy.sh
```diff
@@ -0,0 +1,41 @@
...
+SHA=`git rev-parse --verify HEAD`
+
+# Install sphinx, and build the documentation
+pip install sphinx sphinx_rtd_theme
+cd docs
+make html
+cd ..
+
+# Get ssh credential to push to the documentation, using encrypted key
+openssl aes-256-cbc -K $encrypted_12c8071d2874_key -iv $encrypted_12c8071d2874_iv -in deploy_key.enc -out deploy_key -d
```
RemiLehe: `$encrypted_12c8071d2874_key` and `$encrypted_12c8071d2874_iv` are the secret environment variables that Travis CI maintains.

Issue #358, opened 2019-05-23
Positively charged beam in FBPIC


alexey-petrenko:
Dear Remi, 

I was wondering if it's possible to define a positively charged beam in FBPIC?
I'd like to try FBPIC for some proton-driven PWFA simulations (in the context of AWAKE).

Thanks!

Best regards,
/Alexey

soerenjalas:
We have added this feature. It is implemented in the dev branch and will be available soon with the new release.

PR #364, opened 2019-05-27
Add methods to initialise beams with arbitrary particle charge

soerenjalas:
As requested in #358 with this PR the bunch initialisation methods now accept the arguments `q` and `m` for the charge and mass of the particle species. For backward compatibility I kept the old functions which now just call the new functions with `-e` and `m_e` as arguments.

### DIFF
--- a/fbpic/lpa_utils/bunch.py
+++ b/fbpic/lpa_utils/bunch.py
@@ -13,6 +13,456 @@ from fbpic.particles.elementary_process.cuda_numba_utils import \
 from fbpic.particles.injection import BallisticBeforePlane
 import warnings
 
+def add_particle_bunch( sim, gamma0, q, m, n, p_zmin, p_zmax, p_rmin, p_rmax,
+                p_nr=2, p_nz=2, p_nt=4, dens_func=None, boost=None,
+                direction='forward', z_injection_plane=None ) :
+    """
+    Introduce a simple relativistic electron bunch in the simulation,
+    along with its space charge field.
+
+    Uniform particle distribution with weights according to density function.
+
+    Parameters
+    ----------
+    sim : a Simulation object
+        The structure that contains the simulation.
+
+    gamma0 : float
+        The Lorentz factor of the electrons
+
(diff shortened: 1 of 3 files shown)


fbpic/lpa_utils/bunch.py
```diff
@@ -345,11 +373,11 @@ def add_elec_bunch_openPMD( sim, ts_path, z_off=0., species=None, select=None,
     z = z - np.average(z, weights=w) + z_off
```
soerenjalas: In principle we could also use the charge and mass from the openPMD file here. I just wasn't sure if these properties were always saved and how backward compatible this is.

RemiLehe:
Thanks a lot for this PR! 

I have a few additional suggestions. Let me know what you think (if you agree with them, I could also help implementing this ; let me know what works best):
- For consistency with the `add_new_species` API, I think it would be good if these functions could have `q` and `m` as their first argument (or actually second argument: just after `sim`).
- Also, similarly to the `add_new_species`, it would be nice if the functions could *return* the species object. (e.g. for later use in tracking or diagnostics, I think it is nicer to do `beam.track` rather than `sim.ptcl[-1].track`)
- Since the new functions are more general, I think it would be nice if these are the new recommended way to initialize a beam (even in the case of an electron bream). If you agree, we could edit the documentation and replace the `elec_` function by the new function? We could still add a message saying that the functions with `elec` are still supported, similarly to the message at the end of this page: https://fbpic.github.io/api_reference/diagnostics.html. In that case, we could also use these functions in the example scripts.
- For the Gaussian bunch, should we in fact replace `Q` by `n_physical_particles`, (and `N` by `n_macroparticles`) so that we can also initialize neutral particles (e.g. photons, neutral atoms)?

soerenjalas:
Hi Remi,
I agree with your remarks. It makes sense to define the bunch not with the charge. We were a little hesitant with the long variables names but couldn't come up with a better not confusing naming. 
I didn't yet work on the documentation but I can do that once we have agreed on the PR.

PR #365, opened 2019-05-28
Optimize threads per block pre-settings for CUDA kernels

MKirchen:
This PR optimizes the threads per block used for several CUDA kernels based on empirical performance tests on the P100 and V100 GPU.

- The code is now optimized to run on P100 and V100 GPUs. It is assumed that the values for the P100 GPU are also good for older models such as the K80 GPU. The code selectively uses different TPB values for the V100 GPU.
- 2D kernels are now called with (1, 128) threads per block in x and y by default.
- The deposition kernels are now called with a very small TPB value (between 8 to 32 depending on the particle shape and hardware).
- New environment variables are introduced that can be used to force global TPB values for performance testing. Variables: `FBPIC_1D_TPB`,`FBPIC_2D_TPBX`,`FBPIC_2D_TPBY`

The performance of some kernels was improved by up to a factor of 2. The overall performance of the code on a P100 and V100 GPUs can be expected to be increased by up to ~20% for some cases.

For the tests a 2048x400 box filled with a 16 ppc cold plasma has been used. The tests were performed with 2 and 3 modes and for linear and cubic particle shapes. Advanced features (such as ionization) have not been tested.

The images below show the decrease in cumulative kernel runtime (not all kernels shown) for this PR (orange) compared to the dev branch (blue) for 2 modes and linear particle shapes on a P100 and V100 GPU.

P100

V100


fbpic/particles/particles.py
```diff
@@ -222,6 +222,10 @@ def dens_func( z, r ) ...
             self.prefix_sum_shift = 0
             # Register boolean that records if the particles are sorted or not
             self.sorted = False
+            # Define CUDA threads per block for deposition and gathering kernels
+            deposit_tpb = 16 if cuda_gpu_model == "V100" else 8
+            self.deposit_tpb = 32 if particle_shape == "cubic" else deposit_tpb
+            self.gather_tpb = 256 if particle_shape == "cubic" else 128
```
RemiLehe: While I like the concise one-line syntax `if .. else` for the other cases, here specifically I think that it would be clearer to do:
```
if particle_shape == "cubic":
    self.deposit_tpb = 32
    self.gather_tpb = 256
else:
    self.deposit_tpb = 16 if cuda_gpu_model == "V100" else 8
    self.gather_tpb = 128
```

PR #375, opened 2019-06-22
Very small text fixes

StanczakDominik:
Hi, I'm reading through the code for educational purposes :wink: and there's a couple of very small simplifications, plus a typo fix, that I found could easily be done.


RemiLehe:
Great! Thanks for your interest in the code, and thanks for these edits!

PR #376, opened 2019-06-24
Do not create particle species, when `n_e` is `None`

RemiLehe:
Currently, in FBPIC, when the user does not pass `n_e` (i.e. `n_e` is `None`), no macroparticles are created, but the the electron species itself is still created (and is empty of any macroparticles).

This is not very elegant, and usually forces users to explicitly remove that electron species with:
```
sim.ptcl = []
```

With this PR, no electron species is created, when `n_e` is None. Note that this breaks backward-compatibility in some cases, but these cases are expected to be rare.

Examples where this breaks backward capability:
- The user creates the `Simulation` object with `n_e=None`, forgets to do `sim.ptcl = []`, calls `sim.create_new_species` and then addresses particles (e.g. in `ParticleDiagnostics`) by indexing `sim.ptcl` (e.g. `sim.ptcl[1]`) instead of addressing them with the object returned by `sim.add_new_species`. In this case, this PR changes the indices that should be used (e.g. `sim.ptcl[0]` instead of `sim.ptcl[1]`.
- In the context of ionization: the user creates the `Simulation` object with `n_e=None`, and intentionally omits `sim.ptcl = []` so as to use the empty electron species to store electrons produced by ionization (by passing `sim.ptcl[0]` to `make_ionizable`). With this PR, this empty electron species should be explicitly created with `sim.add_new_species`.

In general, to avoid those issues, the recommended way to create/address particles is to use the object returned by `sim.add_new_species`.

### DIFF
--- a/docs/source/example_input/boosted_frame_script.py
+++ b/docs/source/example_input/boosted_frame_script.py
@@ -18,7 +18,7 @@ where fbpic_object is any of the objects or function of FBPIC.
 # Imports
 # -------
 import numpy as np
-from scipy.constants import c, e, m_e
+from scipy.constants import c, e, m_e, m_p
 # Import the relevant structures in FBPIC
 from fbpic.main import Simulation
 from fbpic.lpa_utils.laser import add_laser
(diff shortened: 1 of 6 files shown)


MKirchen:
Hey, thanks for the PR! Looks good and I am okay to merge.
However, I thought that it would now be the right time to also adjust the example scripts and always use `sim.add_new_species` to create the Particle species explicitly. This will help new users to adopt the newer and better API style.

RemiLehe:
@MKirchen Good point! I just updated the examples and pushed the changes to this PR.

PR #380, opened 2019-07-17
Add note about in-focus laser energy

berceanu:
This PR adds a note to the docs explaining how to relate the laser amplitude which is passed as input to FBPIC to the measured laser energy, both on-target as well as inside the focal spot.

The text is mostly a verbatim copy of @RemiLehe's reply from the official `FBPIC` [slack channel](https://slack-fbpic.herokuapp.com).

### DIFF
--- a/docs/source/api_reference/lpa_utilities/laser.rst
+++ b/docs/source/api_reference/lpa_utilities/laser.rst
@@ -44,6 +44,19 @@ Generic function for arbitrary laser profile
 
 .. autofunction:: fbpic.lpa_utils.laser.add_laser_pulse
 
+.. note::
+    Even though the user does not directly input the laser *energy* into ``FBPIC``,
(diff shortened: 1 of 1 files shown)


MKirchen:
Hey,

thanks for the PR! Although your statements are correct and could be helpful to scientists that are new in the field of plasma acceleration, they are out of scope for the (current) documentation of FBPIC.

I agree that it is important to know that real life laser pulses contain less of their energy in the focal spot compared to their theoretical counterpart. However, this is a purely practical issue and has no direct relation to the properties of the theoretical (but very common) laser profiles available in FBPIC.

Nevertheless, I agree that it would be helpful if we add e.g. the mathematical expressions to calculate the laser energy contained in a given pulse defined with FBPIC in the documentation.

We would be happy to merge a PR that adds such a feature.

I will close the PR. Anyways: Thanks for your work!

berceanu:
Sorry, my bad, should've asked first :)

PR #384, opened 2019-08-09
Add the option `initialize_self_fields`

RemiLehe:
In the PICMI standard, the user can choose whether the initial self fields should be initialized or not.
In order to ensure compatibility, I added the option `initialize_self_fields` to the function that add a bunch to the simulation. (Only the functions with the new API, not the `add_elec_bunch` functions.)

Note that this is not yet fully PICMI-compliant, since a bunch which is loaded with an `AnalyticalDistribution` will not have the option, through FBPIC, to have its space-charge initialized.


MKirchen:
👍

Issue #385, opened 2019-08-14
Using z_injection_plane to "freeze" the beam for a given distance in a non-boosted-frame simulation


abonatto:
Dear Developers,

I was looking for a way to keep the beam "frozen" for a given propagation distance (or time), and I thought z_injection_plane could work. By reading its description,

"z_injection_plane (float (in meters) or None) – When z_injection_plane is not None, then particles have a ballistic motion for z<z_injection_plane. This is sometimes useful in boosted-frame simulations. z_injection_plane is always given in the lab frame."

it seemed to me that it could be used for non-boosted-frame simulations. However, after trying, I saw that this is not the case: 

`AttributeError: 'BallisticBeforePlane' object has no attribute 'inv_gamma_boost'`

Does FBPIC has a way to prevent the beam to evolve for a given distance (or time)?

Regards,

Alex

RemiLehe:
Thanks for reporting this issue!

In principle, `z_injection_plane` should work when running simulations without boosted frame, but you are correct that the current version of FBPIC contains a bug in that case.

I fixed it in PR #386. Once this PR is merged, you could test the bug fix by installing the `dev` version of FBPIC:
```
pip uninstall fbpic  # Uninstall previously-installed version of FBPIC
git clone https://github.com/fbpic/fbpic.git
cd fbpic
python setup.py install
```

With the bug fix, this should allow you to do what you want. Note however that the beam evolves ballistically before the injection plane. This means that the beam momenta are frozen, but that the positions of the particles still evolve. Thus, if the beam has a non-zero emittance, its shape (e.g. transverse RMS size) will still evolve, even before the injection plane. 
Is that acceptable for your use case?

abonatto:
Dear Remi,

Thank you for your support, and for your explanation about the ballistic evolution of the beam. Indeed, I am looking for a way to prevent the transverse RMS size evolution while injecting a beam with non-zero emittance  into the plasma. Do you have any suggestion? Would it be a better approach to set the focus position (zf) to be equal to the beam longitudinal length (6*sigma_z, for example), in such a way the it will be at focus right after fully entering the plasma?

Regards,

Alex

RemiLehe:
Hey, sorry for the late reply (I was away from the office for a few days)!

[...]

PR #386, opened 2019-08-14
Fix bug in ballistic injection for lab-frame simulation

RemiLehe:
As reported in #385, the ballistic injection has a bug when using it in lab-frame simulations.

This pull request fixes the bug ; I checked it locally by adding a gaussian particle beam to the standard LWFA example, with an injection plane.


MKirchen:
Thanks for the fix, okay to merge.

Issue #388, opened 2019-08-24
Boosted-frame simulation of laser-wakefield acceleration - error running boosted_frame_script.py


abonatto:
Dear Developers,

I got the following error when trying to run the script example provided in FBPIC documentation (boosted_frame_script.py):

`Traceback (most recent call last):
  File "boosted_frame_script.py", line 223, in <module>
    sim.step( N_step )
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/fbpic/main.py", line 412, in step
    diag.write( self.iteration )
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/fbpic/openpmd_diag/boosted_particle_diag.py", line 136, in write
    self.store_snapshot_slices(iteration)
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/fbpic/openpmd_diag/boosted_particle_diag.py", line 184, in store_snapshot_slices
    snapshot.prev_z_boost, time, self.select)
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/fbpic/openpmd_diag/boosted_particle_diag.py", line 595, in extract_slice
    species, current_z_boost, previous_z_boost, t )
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/fbpic/openpmd_diag/boosted_particle_diag.py", line 688, in get_particle_data
    pref_sum_curr = pref_sum.getitem( z_cell_curr*(Nr+1) - 1 )
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/numba/cuda/cudadrv/devicearray.py", line 481, in getitem
    return self._do_getitem(item, stream)
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/numba/cuda/cudadrv/devicearray.py", line 490, in _do_getitem
    newdata = self.gpu_data.view(*extents[0])
  File ".../anaconda2/envs/fbpic/lib/python2.7/site-packages/numba/cuda/cudadrv/driver.py", line 1234, in view
    pointer = drvapi.cu_device_ptr(base)
TypeError: int expected instead of float`

Although this is the first time I am running such script, I thought the issue could be related to my recent update of FBPIC (0.13.1) and / openPMD (0.9...). However, I just tested a previous simulation (without laser, in the lab frame), and it worked fine.

Cheers,

Alex

RemiLehe:
Thanks for reporting this issue!

[...]

PR #389, opened 2019-08-25
Fix Python 2 error in particle diag

RemiLehe:
`math.ceil` returns an [integer in Python 3](https://docs.python.org/3.7/library/math.html#math.ceil) but a [float in Python 2](https://docs.python.org/2/library/math.html#math.ceil).

The fact that the returned result is a float caused the boosted-frame diagnostic to fail on GPU: see #388. It is a bit surprising that this error only shows up now, but I am guessing that maybe some of our dependencies were recently updated.

The current PR forces the return result to be an integer.


MKirchen:
Thanks 🙏

PR #392, opened 2019-09-26
Adding an input deck diagnostic

pots007:
This pull request would add an option for the user to store the input deck used to run the simulation as an attribute in hdf5 files via an InputDeckDiagnostic. This could allow for easy checks of exactly what parameters were used in the simulation at a later, when the input deck may have gotten lost. 
Additionally, an arbitrary list of other parameters can be written as attributes into the hdf5 files. For example, if the simulation was part of a large parameter scan (f in np.linspace(1, 3, 0.25)) the user can store the parameter range and the value used in the current simulation. This could simplify automatic analysis later on.

### DIFF
--- a/fbpic/openpmd_diag/inputdeck_diag.py
+++ b/fbpic/openpmd_diag/inputdeck_diag.py
@@ -1,0 +1,120 @@
+"""
+
+inputdeck_diag.py
+
+opmd derived diagnostic for saving the input deck
+into each simulation dump.
+
+"""
+
+import sys
+import os
+import numpy as np
+from fbpic.openpmd_diag.generic_diag import OpenPMDDiagnostic
+
+
+class InputDeckDiagnostic(OpenPMDDiagnostic):
+    """
+    Class that allows saving input decks to dumps.
+    """
+    def __init__(self, period, comm=None, param_dict=None,
+                 write_dir=None, iteration_min=0,
+                 iteration_max=np.inf, dt_period=None, dt_sim=None ):
+        """
+        Setup of the input deck diagnostic
+
+        Parameters
+        ----------
+        period : int, optional
+            The period of the diagnostics, in number of timesteps.
+            (i.e. the diagnostics are written whenever the number
+            of iterations is divisible by `period`). Specify either this or
+            `dt_period`.
+
+        comm : an fbpic BoundaryCommunicator object or None
+            If this is not None, the data is gathered on the first proc
+            Otherwise, each proc writes its own data.
+            (Make sure to use different write_dir in this case.)
+
+        param_dict : a dictionary of additional attributes
+            If this is not None, the dictionary entries will be written
+            to each file dump. The keys of the dict should be in camelCase.
+
+        write_dir : string, optional
+            The POSIX path to the directory where the results are
+            to be written. If none is provided, this will be the path
+            of the current working directory
+
+        iteration_min, iteration_max: ints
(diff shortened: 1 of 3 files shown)


fbpic/openpmd_diag/inputdeck_diag.py
```diff
@@ -0,0 +1,120 @@
+"""
+
+inputdeck_diag.py
+
+opmd derived diagnostic for saving the input deck
+into each simulation dump.
+
+"""
```
RemiLehe: Could you replace the header by the following one:
```
# Copyright 2019, FBPIC contributors
# Authors: Kris Poder
# License: 3-Clause-BSD-LBNL
"""
This file defines the class InputScriptDiagnostic.

It allows saving the input script into each simulation dump.
"""

File: fbpic/openpmd_diag/inputdeck_diag.py [hunk 2]
```diff
@@ -0,0 +1,120 @@
...
+
+import sys
+import os
+import numpy as np
+from fbpic.openpmd_diag.generic_diag import OpenPMDDiagnostic
+
+
+class InputDeckDiagnostic(OpenPMDDiagnostic):
```
RemiLehe: For FBPIC (and Warp/WarpX), we typically use the term "input script", rather than "input deck" (e.g. in documentation)
For this reason could replace "deck" with "script" throughout the file? Also could you rename the current file?
(`git mv inputdeck_diag.py inputscript_diag.py`)

fbpic/openpmd_diag/inputdeck_diag.py
```diff
@@ -0,0 +1,120 @@
...
+        py_files = [f for f in args if '.py' in f]
+        if not py_files:
+            return
+        with open(py_files[0], 'r') as f:
+            text = f.read()
+        return text
+
+    def write_hdf5(self, iteration):
+        """
+            Write an HDF5 file that complies with the OpenPMD standard
```
RemiLehe: Could you reindent the whole text of this docstring (moving it 4 spaces to the left)

fbpic/openpmd_diag/inputdeck_diag.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        # Return right away if no input deck found
+        if self._input_deck is None:
+            return
+
+        filename = "data%08d.h5" % iteration
+        fullpath = os.path.join(self.write_dir, "hdf5", filename)
+
+        # Open the file again, and set the input deck as attribute
+        f = self.open_file(fullpath)
```
RemiLehe: Lines 106 to 116 should be in an `if` statement that makes sure that only the first MPI rank executes them. (In a case of an MPI-parallel simulation). For instance:
```
if self.rank == 0:
```
or
```
if f is not None:
```
(since `self.open_file` returns `None`, for the other MPI ranks.)

fbpic/openpmd_diag/inputdeck_diag.py
pots007: Sorted!

fbpic/openpmd_diag/inputdeck_diag.py
pots007: Sorted!

pots007:
Thanks for the comments, Remi! I've implemented those things. I also updated the __init__.py file to reflect the existence of this new diagnostic.

PR #394, opened 2019-10-20
Close #393: fbpic does not work with new numba release (0.46) on GPU

### LINKED ISSUE #393: fbpic does not work with new numba release (0.46) on GPU
The new release introduced an incompatibility with `cupy`, which will be resolved in the next version of `cupy`.
(The code crashes with 
```
File "cupy/core/core.pyx", line 2519, in cupy.core.core._convert_object_with_cuda_array_interface
TypeError: zip argument #2 must support iteration
```)

We should raise an error in `utils/cuda.py` and recommend that the user downgrades to `numba 0.45`.

RemiLehe:
This PR tells the user to install an earlier version of numba instead.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -11,7 +11,7 @@ This file steers and controls the simulation.
 # as it sets the cuda context)
 from fbpic.utils.mpi import MPI
 # Check if threading is available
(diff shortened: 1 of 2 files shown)


MKirchen:
👍

PR #395, opened 2019-10-20
Update documentation on number of modes

RemiLehe:
The documentation did not explicitly state that one only needs 1 particle in the theta direction when `Nm=1`.

Thanks a lot to @abonatto for pointing this out to me!


MKirchen:
Thanks!

abonatto:
@RemiLehe,

I am the one who thank you and other FBPIC developers. Even being a very small contribution, it feels really good to be able to collaborate!

Cheers!

Alex

PR #397, opened 2019-11-01
Add instructions for Nsight Systems

RemiLehe:
Nvidia released a new profiling tools (Nsight Systems).

This adds instructions on how to profile with NSight Systems.


MKirchen:
🙏

Issue #369, opened 2019-06-05
Improve treatment of high-radius particles


RemiLehe:
Currently, the treatment of high-radius particles (i.e. particles that are outside of the box radially) in FBPIC is not rigorous:
- High-radius particles are kept in the simulation (even though they are not inside the simulation domain anymore)
- They still gather the field, as if they were in the last radial cell.
- They still deposit their charge and current as if they were in the last radial cell ; charge deposited beyond `rmax` (because of the width of the particle shape) is wrapped back into the domain. While wrapping the charge makes sense close to the axis, it does not make much sense at high radius.

It would be good to correct this behavior.

danielseipt:
I have also seen that these high-radius particles can re-enter the simulation box and, for instance, be trapped in a 2nd or 3rd wakefield bucket

danielseipt:
Here is an example input deck that shows the returning particles. In the plot the grey dashed lines mark the boundary of the simulation box.

[bugreport_out_of_box_particles.log](https://github.com/fbpic/fbpic/files/3790051/bugreport_out_of_box_particles.log)


![bugreport_fast_00023750](https://user-images.githubusercontent.com/39627867/67877472-f9784200-fb0f-11e9-8b43-bec342072a85.png)


[bugreport_movie(5).avi.zip](https://github.com/fbpic/fbpic/files/3790151/bugreport_movie.5.avi.zip)

RemiLehe:
Thanks a lot for sending this reproducer!
This should be relatively easy to fix (we should basically prevent the high-radius particles from gathering the fields). I'll try to fix this in the next few days.

RemiLehe:
@danielseipt The issue that you are reporting should now be fixed in the `dev` branch of fbpic.

This fix will be part of the next release. But in the meantime, if you want to use it already, you can install the `dev` branch by doing:
```
pip uninstall fbpic
pip install git+https://github.com/fbpic/fbpic.git@dev
```

MKirchen:
Closed by #398

PR #398, opened 2019-11-04
Do not gather field for the particles that are outside simulation box

RemiLehe:
This PR prevents the high-radius particles (i.e. outside the simulation box) from gathering fields. This fixes a bug in which high-radius particles could get "reflected" into the box (as reported e.g. by @danielseipt in #369). 

This is implemented by adding an `if` condition directly in the gathering kernels. Note that the position of the `if` condition is different in the generic `one_mode` kernels and in the specialized kernels that gather for `m=0` and `m=1` simultaneously. This is because: 
- in the case of the `one_mode` kernels, the fields are set to 0 (after being allocated with arbitrary values) *before* calling the gathering kernels. Therefore, the kernel can be mostly skipped for high-radius particles.
- in the case of the specialized kernels, the fields are not set to 0 beforehand, and therefore, this needs to be done as part of the gathering kernels (even for the high-radius particles)

### DIFF
--- a/fbpic/particles/gathering/cuda_methods_one_mode.py
+++ b/fbpic/particles/gathering/cuda_methods_one_mode.py
@@ -120,84 +120,88 @@ def gather_field_gpu_linear_one_mode(x, y, z,
         # Positions of the particles, in the cell unit
         r_cell =  invdr*(rj - rmin) - 0.5
         z_cell =  invdz*(zj - zmin) - 0.5
-        # Original index of the uppper and lower cell
-        ir_lower = int(math.floor( r_cell ))
-        ir_upper = ir_lower + 1
-        iz_lower = int(math.floor( z_cell ))
-        iz_upper = iz_lower + 1
-        # Linear weight
-        Sr_lower = ir_upper - r_cell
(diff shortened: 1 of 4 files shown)


fbpic/particles/gathering/cuda_methods_one_mode.py
```diff
@@ -120,84 +120,88 @@ def gather_field_gpu_linear_one_mode(x, y, z,
...
-            iz_lower, iz_upper, ir_lower, ir_upper,
-            S_ll, S_lu, S_lg, S_ul, S_uu, S_ug )
-        # Convert to Cartesian coordinates
-        # and write to particle field arrays
-        Bx[i] += cos*Fr - sin*Ft
-        By[i] += sin*Fr + cos*Ft
-        Bz[i] += Fz
+
+        # Only perform gathering for particles that are inside the box radially
+        if r_cell+0.5 < Nr:
```
RemiLehe: The only change here is that I added the `if` condition, and reindented the code.

MKirchen:
Looks good, thanks for the fix!

PR #402, opened 2019-11-12
Close #396: Keep field arrays sorted whenever needed

### LINKED ISSUE #396: Weird electron density structure with ionization turned on
Hi Remi,

As I mentioned to you at APSDPP, I saw weird plasma density structure near the boundary of the box when I use neutral hydrogen as target. Here I attached a snapshot to illustrate that. I plotted the ionized electron density and the laser E fields. It seems that the E fields are fine even if I enhanced the fields by limiting the color scale.
The detailed evolution of the electron density and the input deck can be found in the zipped file.

I need to mention that there wasn't such kind of structure with pre-ionized plasma.

Thanks,
Yong
 
comp7500
[Archive.zip](https://github.com/fbpic/fbpic/files/3790629/Archive.zip)

RemiLehe:
## Overview

During the PIC loop, we typically: 
- gather the E and B fields into the arrays with one element per particle
- then use these fields to push the particles and/or perform ionization

However, when running on GPU, particles arrays may be sorted and rearranged between the time at which the field are gathered and the time at which they are used. (This may happen because of charge/current deposition and/or boosted-frame diagnostics and particle-density diagnostics.) **In the current `dev` branch, when particles are rearranged, the particle arrays for the E and B field are NOT rearranged.** This means that, if the fields are used after the particles have been rearranged, their values will be invalid. (For instance, this caused spurious ionization in #396.)

This PR fixes the issue by rearranging the fields array whenever particles are rearranged. However, because this will impact performance, this is done only when needed i.e. between the gathering step and the particle/push and/or ionization step. In order to implement this, a new flag (`keep_fields_sorted`) is introduced in the `Particles` class. 

In addition, the ionization step has been moved, so that it happens just before the current deposition (instead of just after the current deposition). This is for performance reason: in order to avoid another sort/rearranging call (associated with the current deposition) in the region of the PIC loop where the fields have to be rearranged too.

## Tests

I ran the automated tests on GPU and they passed successfully.

I also tested this PR on the original problem of #396 and it shows that the problem of spurious ionization is fixed.

Screenshot from 2019-11-12 09-01-16

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -400,6 +400,10 @@ class Simulation(object):
             # Main PIC iteration
             # ------------------
 
+            # Keep field arrays sorted throughout gathering+push
+            for species in ptcl:
+                species.keep_fields_sorted = True
+
             # Gather the fields from the grid at t = n dt
             for species in ptcl:
                 species.gather( fld.interp )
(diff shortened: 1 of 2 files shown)


MKirchen:
Thanks and also thanks to @YongMa1988 for reporting this important issue!

PR #403, opened 2019-11-13
Implement few-cycle laser pulse profile

RemiLehe:
When a laser pulse is so short that it contains **only a few laser cycles**, the standard Gaussian profile `GaussianLaser` is not well-adapted, as it does properly take into account spatial-temporal correlations for tightly-focussed pulses. In particular, when initializing a `GaussianLaser` with a short duration **out of focus** in FBPIC, the later profile **at focus** will not be the expected one.

This can be fixed by using the new `FewCycleLaser` introduced by this PR. I also added an automated test that checks that this laser has the proper profile after propagation to focus.

Here is how this pulse looks like:

- out of focus, at the beginning of the test:
FewCycle_before_focus
- at focus, at the end of the test:
Fewcycle_at_focus

By contrast, here is how `GaussianLaser` looks like after propagation to focus, using the same parameters (waist, propagation distance, etc.) as in the above test:
Gaussian_at_focus

PS: This PR is the result of discussions with Julius Huijts (LOA), based on earlier work by Neil Zaim (LOA). Many thanks to both of them!

### DIFF
--- a/fbpic/lpa_utils/laser/laser_profiles.py
+++ b/fbpic/lpa_utils/laser/laser_profiles.py
@@ -728,11 +728,11 @@ class FewCycleLaser( LaserProfile ):
 
         .. math::
 
-            \omega_0 \\tau_{WFHM} = s\\sqrt{2(4^{1/(s+1)}-1)}
+            \omega_0 \\tau_{FWHM} = s\\sqrt{2(4^{1/(s+1)}-1)}
 
         .. note::
 
-            In the case of :math:`\omega_0 \\tau_{WFHM} \gg 1` (i.e. many
+            In the case of :math:`\omega_0 \\tau_{FWHM} \gg 1` (i.e. many
             laser cycles within the envelope), the above expression approaches
             that of a standard Gaussian laser pulse, and thus the :any:`FewCycleLaser`
             profile becomes equivalent to the :any:`GaussianLaser` profile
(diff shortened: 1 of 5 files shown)


fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -693,3 +693,137 @@ def E_field( self, x, y, z, t ):
...
+
+        where :math:`k_0 = 2\pi/\\lambda_0` is the wavevector,
+        :math:`E_0 = m_e c^2 k_0 / q_e` is the field amplitude for :math:`a_0=1`,
+        :math:`Z_R = k_0 w_0^2/2` is the Rayleigh length,
+        :math:`q(z) = z-z_f + iZ_R`, and where :math:`s`
+        controls the duration of the pulse and is given by:
+
+        .. math::
+
+            \omega_0 \\tau_{WFHM} = s\\sqrt{2(4^{1/(s+1)}-1)}
```
MKirchen: Just a typo here and a few lines below: FWHM instead of WFHM

MKirchen:
Thanks for the implementation!

fbpic/lpa_utils/laser/laser_profiles.py
RemiLehe: Good catch, thanks!

PR #404, opened 2019-11-14
Implement Perfectly-Matched-Layers

RemiLehe:
This is the final PR for the implementation of the PML. It includes:
- Implementation of the Maxwell update for the `_pml` fields (standard and galilean)
- Damping of the `_pml` fields (and `Ez` component of the regular fields) in the PML cells (this is done through a new object: `PMLDamper`, which is contained in the boundary communicator object)

In addition, the code for damping+MPI exchanges requires some change in the case of the PML: without PML we used a trick whereby we do **not** perform an inverse Hankel transform before damping/exchanging the fields (see #161). However, this is not possible anymore when PML are used: the fields are damped radially in **interpolation** space, and thus we need to do a full Hankel transform.
In order to handle those different cases, I created a new function `exchange_and_damp_EB` ; this is so that the `step` function remains relatively short and readable.

Finally, I added an automated tests: the test initializes a pulse with a small waist inside the box, and lets it diffract and reach the radial boundary. It then compares the field with the theoretical prediction (which assumes infinite space).

Here is the result of the test **in the absence of PML** (reflections clearly dominate):
Figure_3

Here is the result of the test **in the presence of PML** (reflections are eliminated):
Figure_1


fbpic/boundaries/boundary_communicator.py
```diff
@@ -792,7 +799,7 @@ def exchange_particles_aperiodic_subdomain(self, species, fld, time ):
     def damp_EB_open_boundary( self, interp ):
```
MKirchen: There is a bug in line 855. You pass self.d_left_damp instead of self.d_right_damp to the GPU kernel. Apparently I missed this in the previous PR.

fbpic/boundaries/boundary_communicator.py
```diff
@@ -47,7 +48,7 @@ class BoundaryCommunicator(object):
 
     def __init__( self, Nz, zmin, zmax, Nr, rmax, Nm, dt, v_comoving,
             use_galilean, boundaries, r_boundary, n_order, n_guard,
-            n_damp, nr_damp, n_inject=None, exchange_period=None,
+            n_damp, nr_damp, cdt_over_dr, n_inject=None, exchange_period=None,
```
MKirchen: The parameter cdt_over_dr is missing in the docstring

PR #409, opened 2019-11-24
Suggestion for a 3D plasma density profile option

hightower8083:
This PR adds a  3D plasma density profile functionality, which we use in our simulations from time to time, and I think it could be useful for others. 

With this PR one can provide a `dens_func` to be function of `(x, y, z)` or `(z, r, theta)`, which has to be specified as a doc-string  `dens_func.__doc__` being either `"xyz"` or `"zrt"`. Any other value of `dens_func.__doc__` leads to the default `(z, r)` behaviour.

Additionally, there is a `density_profiles.py` module, where  class `DensityProfileFromPolarGrid` can construct from a transverse profile defined as an on-grid input `(r, theta)` into a function truncated  to a limited number of azimuthal modes. This can improve numerical stability in the particular cases of _boosted_-simulations where axial asymmetries cause noise excitation. This example is present here just for consideration (not for merging) -- if there is an interest it will be presented in a separate PR.

### DIFF
--- a/fbpic/particles/injection/continuous_injection.py
+++ b/fbpic/particles/injection/continuous_injection.py
@@ -8,6 +8,14 @@ It defines a class for continuous particle injection with a moving window.
 import warnings
 import numpy as np
 from scipy.constants import c
+import sys, inspect
+
+# convenience method retrieve arguments number of a function
+if sys.version_info[0] < 3:
+    get_args_len = lambda fu: len(inspect.getargspec(fu)[0])
+else:
+    get_args_len = lambda fu: len(inspect.signature(fu).parameters)
+
 
 class ContinuousInjector( object ):
     """
@@ -38,6 +46,14 @@ class ContinuousInjector( object ):
         self.uy_th = uy_th
         self.uz_th = uz_th
 
+        # Define and register dimensions number for density profile
+        if self.dens_func is not None:
+            self.dens_func_dim = get_args_len(self.dens_func)
+            if self.dens_func_dim not in (2, 3):
+                raise ValueError(
+                "Density function can have 2 or 3 arguments, " + \
+                "but got {:d} instead".format(self.dens_func_dim) )
+
         # Register spacing between evenly-spaced particles in z
         if Npz != 0:
             self.dz_particles = (zmax - zmin)/Npz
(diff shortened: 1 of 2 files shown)


fbpic/particles/injection/continuous_injection.py
```diff
@@ -168,8 +167,17 @@ def generate_particles( self, time ):
...
-                return( self.dens_func( z-self.v_end_plasma*time, r ) )
+            if self.dens_func.__doc__ == 'xyz':
+                def dens_func( x, y, z ):
+                    return( self.dens_func(x, y, z-self.v_end_plasma*time ) )
+            elif self.dens_func.__doc__ == 'zrt':
+                def dens_func( z, r, th ):
+                    return( self.dens_func(z-self.v_end_plasma*time, r, th ) )
+            else:
+                def dens_func( z, r ):
+                    return( self.dens_func( z-self.v_end_plasma*time, r ) )
```
RemiLehe: Cool idea!
However, instead of using `__doc__`, could we check the names in `self.dens_func.__code__.co_varnames`?

hightower8083: well, that's  indeed a more clear way! 
The drawback is that it won't do when function is defined as a class with the `__call__` method (as it's done in `density_profiles.py` and laser profiles). In this case `dens_func` won't have `__code__` attribute..

On one hand, I can adapt the behaviour of `density_profiles.py`, but I though that making `dens_func` as a class (like laser profile) could be an option for future development..

hightower8083: @RemiLehe whenever you have time, let me know which way is preferable -- for me personally, setting `dens_func.__doc__='xzy'` in the input is a more explicit (=deliberate) way of doing things, but if you prefer to have it as an out-of-box feature, I'll adjust the PR

MKirchen:
Hey @hightower8083, thanks for this PR! @RemiLehe and I discussed this topic offline and we both agree that we have to improve the API of defining the plasma density profile as you suggest in this PR. Already for a long time we want to introduce objects (similar to the laser profiles) to define the plasma profile and remove the current API of just passing a `dens_func` (The second part of your PR goes into this direction).

Although your proposed changes of the first part of the PR would make immediate improvements possible, I think that defining the coordinate system by setting `dens_func.__doc__` is too unusual and not very pythonic. We decided that we would rather not merge this as is.

We will discuss a potential new API in the next few weeks again in detail and come back to you.

hightower8083:
OK, let me know

PR #410, opened 2019-11-26
Fix formula to calculate speedup in the boosted frame

MKirchen:
This PR fixes a mistake in the Lorentz-boosted frame documentation. The formula for the scaling law of the speedup in the boosted frame was wrong in the text and in the figure. 

For reference, the two notebooks that were used to create the figures of the speedup and the resolution in boosted frame are attached to this PR.

[Boosted_frame_docs_notebooks.zip](https://github.com/fbpic/fbpic/files/3891367/Boosted_frame_docs_notebooks.zip)


RemiLehe:
Thanks for fixing this!

PR #411, opened 2019-11-26
Shift CEP phase of FlattenedGaussian Laser by -pi/2

MKirchen:
This PR fixes the default CEP phase of the `FlattenedGaussianLaser` profile.

Although the CEP phase is 0 by default, the field of the `FlattenedGaussianLaser` profile would be 0 at the peak of the envelope. In contrast, the field of the `GaussianLaser`, for example, would be at the maximum, which is the correct and expected behaviour.

This PR shifts the CEP phase by `-pi/2` in the definition of the FlattenedGaussianLaser profile, such that the field is at its maximum for `cep_phase=0`.


RemiLehe:
Thanks!

PR #415, opened 2019-12-09
Update the API of the PML

RemiLehe:
This updates the API for the PML:
- `boundaries` should now be passed as a dictionary ; e.g. `boundaries = {'z':'open', 'r':'reflective'}` without PML and `boundaries = {'z':'open', 'r':'open'}` for PML.
- `n_damp` should also be passed as a dictionary (although users usually don't pass this, and use the default instead).
Note that the previous API (single string. e.g.`boundaries='open'`) still works (for backward compatibility) but will raise a warning, recommending to pass a dictionary instead.

A few minor changes related to the API were also added:
- The PICMI interface can now also activate the PML. (by passing `'open'` for the `r` boundary)
- The `r` boundary (`'open'` or `'reflective'`) is printed when doing `verbose_level=2`.
- The default example scripts now use the new API


MKirchen:
Thanks!

PR #416, opened 2019-12-10
Fix bug in GPU-MPI communications of PML

RemiLehe:
The communications of PML fields has a typo, in the case where the code runs on GPU.

This PR fixes the typo. The code can be tested by doing:
```
mpirun -np 2 python test_pml.py
```
(and changing `n_order` to `32` in `test_pml.py`)

Many thanks to @Kookjin for testing the `finalized_pml` branch and reporting an issue in GPU-MPI simulations!! :sparkles:


MKirchen:
Good catch!

PR #423, opened 2020-01-13
Use device version of PSATD coefficients for PML push

RemiLehe:
When profiling a run with PML on GPU today, I noticed many Host-Device copies. 

It turns out that I was using the CPU arrays for the PSATD coefficients in the PML push kernel.

This PR fixes the issue ; I checked that this removes the Host-Device copies.


MKirchen:
Thanks for the fix!

PR #424, opened 2020-01-13
Gather only for particles that are inside the physical box

RemiLehe:
This PR is very similar to #398 ("Do not gather field for the particles that are outside simulation box"), but while #398 prevents particles from gathering fields *when they are outside the full simulation grid (which includes PML cells)*, the current PR prevents particles from gathering fields *when they are outside the physical domain (which excludes PML cells)*.

Thus, with this PR, particles will not gather fields from the PML, and will thus propagate in straight lines.

In practice, this PR modifies the code in the same places that #398 did, but checks if `r < rmax_gather` where `rmax_gather` is given by `comm.get_rmax( with_damp=False )` (`with_damp=False` excludes the PML) - see the file `particles.py`. As consequence, the code has been modified so that the `BoundaryCommunicator` `comm` is passed to the `Particles.gather` function.


MKirchen:
Thanks!

PR #425, opened 2020-01-15
[WIP] Use `cupy` more extensively

RemiLehe:
# Overview

This branch tries to use `cupy` more extensively, in order to take advantage of:
- faster memory management / deallocation of arrays
- lower kernel launch overhead
compared to `numba.cuda`.

In particular, all of the arrays are now allocated with `cupy` instead of `numba.cuda`.

# List of changes

- All of the device arrays are now `cupy` arrays instead of `numba.cuda` array. In order to change this throughout the code, I wrote, committed and executed a Python script. (see the first two commits)
- On top of this, I tried to reduce kernel load overhead by avoiding some `cuda.numba` functions where possible.
(see the commits "Use cupy.take for particle rearranging", "Use Nd FFT to avoid 2d/1d copies"). There are certain more places where we could do this.
- The last-but-one commit is to avoid errors when the data is on the CPU (e.g. before we call `step`) but we already call `deposit` (e.g. when doing the initial space-charge calculation)

# Caveat

All the tests now pass on GPU, except for the laser antenna tests ; the cde for the laser antenna needs some minor synthactic changes in order to fix this.

### DIFF
--- a/fbpic/boundaries/boundary_communicator.py
+++ b/fbpic/boundaries/boundary_communicator.py
@@ -18,7 +18,9 @@ from .pml_damping import PMLDamper
 from .particle_buffer_handling import remove_outside_particles, \
      add_buffers_to_particles, shift_particles_periodic_subdomain
 # Check if CUDA is available, then import CUDA functions
-from fbpic.utils.cuda import cuda_installed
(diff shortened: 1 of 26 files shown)


RemiLehe:
This PR will be replaced by several small PRs, made to the branch `cupy`.

PR #426, opened 2020-01-22
Example for 3D visualization with PyVista in the Docs

lauridsj:
I added a simple example in the documentation on how to create a 3D visualization of simulation results using PyVista (a python wrapper for VTK). The example produces a simple 3D image (volume plot) from the example LWFA script, but could be extended to produce e.g. 3D videos. It should also easily generalize to other cases.


RemiLehe:
Thanks a lot for this PR! The 3D image looks awesome! :star2: 

One minor request: could you actually move the new section ("3D visualization using PyVista") in the section "Advanced use" (https://fbpic.github.io/advanced/advanced.html) ? This is in order to keep the section "How to run the code" short and having it contain only the minimal information to get started.

You can do the above by creating a new file in the folder `advanced/` and listing it in `advanced.rst`. I would be also great to put a link to that new section in `how_to_run.rst`. (i.e. having a only link instead of having the whole section directly in this file.)

Thanks again!

hightower8083:
wow, i didn't know that `theta=None` for `get_field` now returns a full 3D -- cool feature! 

I've noticed, however, that it provides a _full_   3D grid `(Nz, Nr, Nr)`, which  can be very expensive, especially when working with video. Just to mention, there is an alternative way used in [opmd2VTK](https://github.com/hightower8083/opmd2VTK/tree/dev), where one constructs a vtk object based on an _UnstructuredGrid_, with a given number on `N_theta` slices (can also be sub-sampled). Such grid is kinda heavier to play with than the `UniformGrid` (at least in Paraview), but is faster in producing the VTK files, which are also smaller in size.

lauridsj:
@RemiLehe Thanks!

I moved the example to a seperate file and added a link - it should be more clear now.

@hightower8083 Oh, I did not know that package, thanks for the input! It is in fact quite slow when you try to create a movie; I actually went ahead and slightly modified the openPMD code to only give a small part of the grid on my computer.

Unfortunately, PyVista only accepts `UniformGrid` for volume plots (see [here](https://docs.pyvista.org/plotting/plotting.html#pyvista.BasePlotter.add_volume)), so one would have to use a different wrapper or interface with VTK directly. I've not really looked into that, so I probably won't do that in the near future.

RemiLehe:
Thanks for updating the doc!

hightower8083:
@lauridsj, if you are going for modifications of openPMD_viewer you may try to sub-step the grid instead of taking a limited part of it -- looking at plasma density, you may easily take big steps along Z-axis (e.g. [::10]), and something like [::2] in the transverse directions.

In any case, for the fine-tuned animations its often necessary to do rendering with some dedicated tools like Paraview/VisIt, and using series of pre-created vtk files. This tools can work with most of grid types. In its turn, this pyvista lib looks very useful for an interactive work 🤓

PR #429, opened 2020-02-20
Implement context manager to move data to GPU

RemiLehe:
This is a PR that **prepares** the code for a switch of some functionalities from `numba.cuda` to `cupy`.

More specifically, in FBPIC, we sometimes pass CPU (`numpy`) arrays to `numba.cuda` kernels. This essentially happens when we call `deposit` outside of the main PIC loop. (e.g. when computing the initial space-charge fields). In this case, the code calls the `numba.cuda` kernels for charge/current deposition, but the arrays are still on the CPU, because the execution has not gone through the set of `send_to_gpu` functions that are the beginning of the main PIC loop.

In this case, `numba.cuda` kernels are quite forgiving, as they then automatically copy the `numpy` arrays to GPU before the kernel and back to CPU after the kernel.
However, this will not be the case with `cupy` kernels: instead (from experience) the code crashes when passing `numpy` array to `cupy` kernel.

In order to avoid this issue in upcoming PRs with more `cupy` kernels, this PR creates a context manager that moves the simulation data to GPU (and back) if it is originally on CPU. This context manager is used for `deposit`, in this PR, but could also be used in other parts of the code if the need arises.

To keep track of whether the data is on CPU or GPU at any point in time, new boolean flags `data_is_on_gpu` have been added to the `Particles` and `Fields` class. These flags are of course updated when calling `send_to_gpu` and `receive_from_gpu`

### DIFF
--- a/fbpic/fields/fields.py
+++ b/fbpic/fields/fields.py
@@ -136,6 +136,7 @@ class Fields(object) :
                 'Cuda not available for the fields.\n'
                 'Performing the field operations on the CPU.' )
             self.use_cuda = False
+        self.data_is_on_gpu = False # Data is initialized on CPU
 
         # Register the current correction type
         if current_correction in ['curl-free', 'cross-deposition']:
@@ -219,6 +220,7 @@ class Fields(object) :
             for m in range(self.Nm) :
                 self.interp[m].send_fields_to_gpu()
                 self.spect[m].send_fields_to_gpu()
+            self.data_is_on_gpu = True
 
     def receive_fields_from_gpu( self ):
         """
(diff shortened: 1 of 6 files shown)


fbpic/main.py
```diff
@@ -581,61 +582,65 @@ def deposit( self, fieldtype, exchange=False,
...
-        if update_spectral:
-            fld.interp2spect( fieldtype )
-            if self.filter_currents:
-                fld.filter_spect( fieldtype )
-            # Set the flag to indicate whether these fields have been exchanged
-            fld.exchanged_source[ fieldtype ] = exchange
+        # If the simulation data is initially on CPU, move it to GPU
+        # for this function, and then move it back to CPU at the end
+        with MoveSimToGpuIfNeeded(self):
+
```
RemiLehe: The only change in the code below is the indentation of the code.

RemiLehe:
Note that I checked that all of our tests pass on GPU, with this branch.

RemiLehe:
Thanks! I just checked that the automated tests pass on GPU ; I think that we can merge.

PR #431, opened 2020-02-27
Use a custom decorator to compile CUDA kernels to cupy

lauridsj:
This pull request implements a custom decorator mimicking Numba's `@cuda.jit` which compiles python functions to CUDA kernels. It uses the Numba JIT compilation to produce PTX code, which is then passed to Cupy to create a RawKernel. This way, the smaller API overhead from Cupy is utilized while retaining the flexibility of Numba JIT compilation.

### DIFF
--- a/fbpic/boundaries/cuda_methods.py
+++ b/fbpic/boundaries/cuda_methods.py
@@ -6,8 +6,9 @@ This file is part of the Fourier-Bessel Particle-In-Cell code (FB-PIC)
 It defines a set of generic functions that operate on a GPU.
 """
 from numba import cuda
+from fbpic.utils.cuda import compile_cupy
 
-@cuda.jit
+@compile_cupy
 def copy_vec_to_gpu_buffer( vec_buffer_l, vec_buffer_r,
                             grid_r, grid_t, grid_z, m,
                             copy_left, copy_right, nz_start, nz_end ):
@@ -68,7 +69,7 @@ def copy_vec_to_gpu_buffer( vec_buffer_l, vec_buffer_r,
                 vec_buffer_r[3*m+2, iz, ir] = grid_z[ iz_right, ir ]
 
 
-@cuda.jit
+@compile_cupy
 def copy_pml_to_gpu_buffer( vec_buffer_l, vec_buffer_r,
                             grid_r, grid_t, grid_z, pml_r, pml_t, m,
                             copy_left, copy_right, nz_start, nz_end ):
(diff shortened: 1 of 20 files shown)


RemiLehe:
Thanks a lot for this PR!! This is awesome!

A few preliminary remarks (I have not looked at the code in details yet):

- It looks like like the Python 2 tests are [failing](https://travis-ci.org/fbpic/fbpic/jobs/655961797?utm_medium=notification&utm_source=github_status) ; could you check whether there is an easy fix? I know that Python 2 is principle deprecated now ; but if we can still maintain compatibility without too much effort, it is probably worth it.
- Could you confirm that all the GPU tests are passing for this PR? Also, do you know if this works with `cupy` version 7?
- Could you post the benchmark numbers for the standard `lwfa_script.py`?

lauridsj:
Sorry I couldn't reply earlier - I was away for the weekend.

- All tests pass on the GPU independently of the Cupy version.
- The standard `lwfa_script.py `(Nz 800, Nr 50) runs with 25 ms/step on average (17 ms/step on fast iterations, longer on exchange periods). To compare: the current dev branch needs 116 ms/step on average (102 ms/step on fast iterations).
- I'll look into Python 2. I think the only problem is the tuple unpacking (`*a.shape`), so it should be an easy fix.

lauridsj:
I tried Python 2 on GPU and cannot get it to work even with standard Cupy. Note that Cupy 7 is not officially supported for Python 2 (there is only a beta version).

I can try again with older versions, but honestly I'm not sure whether supporting Python 2 for GPU still makes sense.

lauridsj:
Update: With Numba 0.45.0 and Cupy 6.0.0, the dev branch runs on Python 2. However, this branch gives a cryptic CUDA error: `Call to cuPointerGetAttribute results in CUDA_ERROR_INVALID_VALUE`.

Again: Not sure whether it's worth investigating this.

lauridsj:
Here is a snippet from the Cupy changelog that Python 2 support is dropped from Cupy 7.0.0 onwards:
https://github.com/cupy/cupy/releases/tag/v7.0.0

RemiLehe:
OK, thanks a lot for the additional information.
Yes, I agree that we should drop support for Python 2 when running on GPU. I'll make a separate PR that checks that the Python version is 3, when `use_cuda` is True.

RemiLehe:
Thanks a lot for this PR! The code is very well explained!
I pushed some minor modification to the docstrings, but did not modify the code itself.

PR #432, opened 2020-03-04
For GPU: check cupy, numba and Python version.

RemiLehe:
In previous versions of FBPIC, we were allowing `numba<= 0.45` on GPU - provided that `cupy<=6` was also installed. 

However, now that GPU arrays are `cupy` arrays, this is no longer possible. This is because passing `cupy` arrays of size 0 to a `numba.cuda` kernel (which happens e.g. when a given species has 0 macroparticles in FBPIC) causes the code to crash, with `cupy<=6`. (See [this issue](https://github.com/numba/numba/issues/4140) for more details.)

Note that this is still the case with the new decorator (PR #431), since we still pass cupy arrays (potentially of size 0) to `numba`'s `specialize` function - which is enough to cause the above-mentioned crash.

As a result, we now need `cupy>=7`, which in turn imposes `numba>=0.46` and Python 3 (when running on GPU).

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -20,6 +20,7 @@ if cuda_installed:
     mpi_select_gpus( MPI )
 
 # Import the rest of the requirements
+import sys
 import warnings
 import numba
 import numpy as np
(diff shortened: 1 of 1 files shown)


MKirchen:
👍

PR #433, opened 2020-03-05
Move the cumsum in the ionization kernels to the GPU

lauridsj:
This PR moves the cumsum calculated after the particle ionization to the GPU using `cupy.cumsum`. This helps reduce overhead since much less data needs to be copied from device to host and back.

### DIFF
--- a/fbpic/particles/elementary_process/ionization/ionizer.py
+++ b/fbpic/particles/elementary_process/ionization/ionizer.py
@@ -269,16 +269,20 @@ class Ionizer(object):
 
         # Count the total number of new electrons (operation always performed
         # on the CPU, as this is typically difficult on the GPU)
-        if use_cuda:
-            n_ionized = n_ionized.get()
-        cumulative_n_ionized = perform_cumsum_2d( n_ionized )
+        #if use_cuda:
+        #    n_ionized = n_ionized.get()
+        cumulative_n_ionized = perform_cumsum_2d( n_ionized, use_cuda )
         # If no new particle was created, skip the rest of this function
-        if np.all( cumulative_n_ionized[:,-1] == 0 ):
-            return
+        if use_cuda:
+            if cupy.all( cumulative_n_ionized[:,-1] == 0 ):
+                return
+        else:
+            if np.all( cumulative_n_ionized[:,-1] == 0 ):
+                return
         # Copy the cumulated number of electrons back on GPU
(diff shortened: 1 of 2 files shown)


lauridsj:
It should be noted that there are still two very small `memcpy` calls per timestep (both device to host): 

image

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -267,18 +267,15 @@ def handle_ionization( self, ion ):
...
+        # Count the total number of new electrons 
+        cumulative_n_ionized = perform_cumsum_2d( n_ionized, use_cuda )
         # If no new particle was created, skip the rest of this function
-        if np.all( cumulative_n_ionized[:,-1] == 0 ):
-            return
-        # Copy the cumulated number of electrons back on GPU
-        # (Keep a copy on the CPU)
         if use_cuda:
-            d_cumulative_n_ionized = cupy.asarray( cumulative_n_ionized )
+            if cupy.all( cumulative_n_ionized[:,-1] == 0 ):
```
lauridsj: The first memcpy stems from here - `cupy.all` runs on the GPU (as it has to), the resulting `bool` value thus has to be copied to the CPU in order for Python to do the `if` check.

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -288,13 +285,14 @@ def handle_ionization( self, ion ):
         assert len(self.target_species) == n_levels
         for i_level, elec in enumerate(self.target_species):
             old_Ntot = elec.Ntot
-            new_Ntot = old_Ntot + cumulative_n_ionized[i_level,-1]
+            # Cast to int transfers the data from the GPU if needed
+            new_Ntot = old_Ntot + int( cumulative_n_ionized[i_level,-1] )
```
lauridsj: The second memcpy is here - the new number of particles `new_Ntot` depends on the result of the `cumsum`, now on the GPU. It has to be copied to the CPU in order to allocate the new arrays, which is done by the cast to `int`.

lauridsj:
If someone has great ideas on how to get rid of those memcpys, please feel free to share them :)

Even so the cumsum gives a slight speedup, so I would keep it on the GPU anyway.

RemiLehe:

PR #434, opened 2020-03-10
Moved cumsum to GPU in Compton kernels

lauridsj:
This moves the calculation of the cumsum to the GPU in the Compton scattering module, analogously to the ionization module.


RemiLehe:
Thanks! It looks like `pyflakes` is throwing an error - could you have a quick look?

PR #435, opened 2020-03-19
Added call to clear cupy memory pool for each exchange period

lauridsj:
When allocating data on the GPU, cupy by default does not free memory when it is not needed anymore, instead caching the memory so that future allocations become faster. This is expected behaviour and does not lead to out-of-memory crashes since cupy detects when the memory pool is full and clears it when needed. 

However, this leads to the GPU memory filling up even for small simulations, which might be confusing to users. Additionally, it might conflict with other GPU libraries which are not aware of cupy. This PR forces cupy to clear its memory pool at each exchange period to remedy this.

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -18,6 +18,8 @@ if cuda_installed:
     from .utils.cuda import send_data_to_gpu, \
                 receive_data_from_gpu, mpi_select_gpus
     mpi_select_gpus( MPI )
+    if cupy_installed:
+        import cupy
 
 # Import the rest of the requirements
 import sys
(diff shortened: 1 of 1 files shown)


hightower8083:
I've just ran a check and the simulation of ~6Gb has time to rise to 10Gb between exchanges. May be it could be some smaller simpler period,  e.g. each 4 steps? it would still be not as a burden as "each step" way, but will limit better  the memory variations

lauridsj:
The free_all_blocks call does have some significant overhead in smaller simulations, so I'm not sure if calling it more often is really a good idea. Again, this should not lead to OOM crashes since cupy automatically clears its pool if required.

lauridsj:
grafik

This is the memory usage for a simple ionization case. With free_all_blocks, the memory rises to ~150% of the dev branch, which I think is acceptable.

lauridsj:
Regarding runtime, the cupy branch with `free_all_blocks` each exchange period took 83 ms on average while the standard cupy branch took 79 ms. To compare, the dev branch took 711 ms.

PR #438, opened 2020-04-06
PICMI: Add uniform distribution and default number of steps

RemiLehe:
This PR adds support for PICMI's `UniformDistribution` in FBPIC. (Before this PR, the only other supported distribution for evenly-spaced particles is the `AnalyticalDistribution`.)

In addition, this PR incorporates a small change to the `step` function: this is because picmi allows the number of timesteps to be passed either to `step`, or the the `Simulation` `__init__` function.


MKirchen:
👍

PR #440, opened 2020-04-14
Fix GPUdirect for cupy arrays

lauridsj:
Since the last version of FBPIC, GPUdirect (i.e. direct MPI communication between GPUs) no longer works because the `numba` arrays have been replaced by `cupy` arrays. This PR fixes this issue.

It should be noted that GPUdirect only works with the development version of `mpi4py` (3.1.0a0), where a new interface for this functionality has been introduced (see [the cupy documentation](https://docs-cupy.chainer.org/en/stable/reference/interoperability.html)).

The PR also updates the documentation regarding installation on JUWELS to include GPUdirect.

### DIFF
--- a/docs/source/install/install_juwels.rst
+++ b/docs/source/install/install_juwels.rst
@@ -19,7 +19,8 @@ To load the standard modules the ``.bashrc`` should contain the following:
 
 ::
 
-    module load intel-para/2019a
+    module load GCC/8.3.0
+    module load MVAPICH2/2.3.3-GDR
 
 Installation of Anaconda
 ------------------------------------------------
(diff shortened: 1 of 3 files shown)


RemiLehe:
Thanks for this PR!
Could you add the following check in `utils/mpi4py`:
```python
if gpudirect_enabled:
    mpi4py_version_number = mpi4py.__version__.split('.')
    mpi4py_major_version = int(version_number[0])
    mpi4py_minor_version = int(version_number[1])
    if (mpi4py_major_version < 3) or (mpi4py_minor_version < 1):
        raise RuntimeError(
            "In order to use GPU Direct, you need to install mpi4py>=3.1.")
```

lauridsj:
Thanks for your comment!

I added the check for the mpi4py version.

RemiLehe:
:+1:

PR #441, opened 2020-04-17
make setup.py executable

skuschel:


RemiLehe:
Thanks for this PR!

Could you say a few words about the motivation for having the `setup.py` executable?
Do you have use cases where executing `setup.py` directly would be useful?

skuschel:
Hi Remi,
of course! I personally use it a lot to install and rebuild `./setup.py develop --user` the current development version from the git repo. Also `./setup.py test` works to runs the tests. Of course its just a convenience over `python setup.py develop --user`.

RemiLehe:
OK, sounds good!
Thanks for this PR ✨

PR #442, opened 2020-04-19
Move shape factors to seperate file

lauridsj:
In the current code, the particle shape factors are defined redundantly at three seperate locations. This PR moves them to a single file and imports them where needed.


RemiLehe:
Thanks for implementing this!
Just to check: could you confirm that the automated tests pass on GPU?

lauridsj:
Yup, all the tests passon GPU.

RemiLehe:
Thanks! 🙏

PR #452, opened 2020-06-01
Refactor the code for Ruyten correction

RemiLehe:
This refactors the code for the Ruyten correction, by removing the `if` condition in the deposition kernels and instead add an element in the Ruyten array (which corresponds to the particles in the first half of the first cell, radially.)

This should be equivalent to the previous implementation - except for the laser antenna: it seems that the laser antenna was using a non-zero Ruyten coeffficient for the particles in the first half of the first cell in the previous code.

### DIFF
--- a/fbpic/particles/utilities/utility_methods.py
+++ b/fbpic/particles/utilities/utility_methods.py
@@ -78,8 +78,8 @@ def weights(x, invdx, offset, Nx, direction, shape_order, beta_n):
             S[1,:] = 1 - S[0,:]
         # Linear weight r
         elif direction == 'r':
-            ir = np.minimum(np.maximum(i[0,:], 0), Nx-1)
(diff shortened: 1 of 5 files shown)


fbpic/particles/utilities/utility_methods.py
```diff
@@ -78,8 +78,8 @@ def weights(x, invdx, offset, Nx, direction, shape_order, beta_n):
             S[1,:] = 1 - S[0,:]
         # Linear weight r
         elif direction == 'r':
-            ir = np.minimum(np.maximum(i[0,:], 0), Nx-1)
-            S[0,:] = (i[1,:] - x_cell) * (1.+beta_n[ir]*( x_cell - i[0,:] ))
```
RemiLehe, line 82: Note that the above code means that particles located in the first half of the first cell were using a non-zero Ruyten coefficient (`beta_n[0]`, before a zero-element was added to the `beta_n` array)

PR #461, opened 2020-07-09
inactive mode for absent antennas 

hightower8083:
When antenna is not in the simulation domain it still updates velocities of virtual particles, and wastes a bit of resources. 
Here an inactive mode is added when antenna is not in the simulation domain.

### DIFF
--- a/fbpic/lpa_utils/laser/antenna_injection.py
+++ b/fbpic/lpa_utils/laser/antenna_injection.py
@@ -161,7 +161,6 @@ class LaserAntenna( object ):
         # Register whether the antenna deposits on the local domain
         # (gets updated by `update_current_rank`)
         self.deposit_on_this_rank = False
-        self.active_update_v = False
 
         # Initialize small-size buffers where the particles charge and currents
         # will be deposited before being added to the regular, large-size array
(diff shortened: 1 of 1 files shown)


fbpic/lpa_utils/laser/antenna_injection.py
```diff
@@ -239,6 +247,11 @@ def update_v( self, t ):
         t: float (seconds)
             The time at which to calculate the velocities
         """
+        # Interrupt this function if the antenna is not currently
+        # active on the global domain (as determined by `update_current_rank`)
+        if not self.active_update_v:
+            return
+
```
RemiLehe, line 254:
```suggestion
        # Interrupt this function if the antenna is not currently
        # active on the global domain
        zmin_global, zmax_global = comm.get_zmin_zmax(
            local=False, with_damp=True, with_guard=True )
        if (z_antenna >= zmin_global) and (z_antenna < zmax_global):
            return

```

hightower8083: well this would `return` for antenna in domain =)

```suggestion
        # Interrupt this function if the antenna is not currently
        # active on the global domain
        zmin_global, zmax_global = comm.get_zmin_zmax(
            local=False, with_damp=True, with_guard=True )
        if (z_antenna < zmin_global) or (z_antenna > zmax_global):
            return
```

RemiLehe:
Actually, now I see that my advice was quite bad, since `update_x` does not have access to the `comm` object. 
Thus, I agree with your original implementation  ; I'll revert the changes.

PR #470, opened 2020-08-17
Improve performance of external fields on GPU

lauridsj:
This PR improves the performance of external fields when on GPU.

In the current code, `numba.vectorize` is used to compile the user-defined field functions for GPU. This is slow and also causes `memcopy` operations every call. In this PR, `cuda.jit` is instead used to compile an inline function, which is then embedded in a standard kernel compiled with `compile_cupy` to benefit from cupy's smaller kernel call overhead.


RemiLehe:
Thanks a lot for this PR! 🙏

PR #473, opened 2020-08-20
Calculate laser antenna on the GPU

lauridsj:
**This PR is work-in-progress.**

This PR moves the calculation of the laser antenna to the GPU in order to prevent `memcopy` operations every iteration. It also replaces the specialized antenna deposition kernels by the standard kernels. Since the antenna particles are unsorted, for the GPU this neccessitates deposition kernels for unsorted particles.


lauridsj:
All tests pass on GPU. The Gaussian and Few Cycle lasers are GPU compatible. I still have to do some benchmarks whether there is a speedup.

lauridsj:
I did some benchmarks and in my test cases, there was no real time difference between this PR and the old code (neither for a Gaussian nor a Flattened Gaussian laser). I did confirm that for a GPU-capable laser like the Gaussian, there are no memcpys during the antenna calculation.

I think we should merge the PR nonetheless for improved code quality.

fbpic/lpa_utils/laser/antenna_injection.py
```diff
@@ -50,7 +53,8 @@ class LaserAntenna( object ):
     rest of the simulation uses cubic shape factors.)
     """
```
RemiLehe, line 54: Could you update the 3rd paragraph of the docstring of the `LaserAntenna` object above?
More specifically, we should specify that, when running on GPU:
- The update of the velocity of the virtual particles is either done on CPU or GPU, depending on whether the laser profile is GPU enabled.
- The deposition of charge/current is then always done on GPU.

fbpic/lpa_utils/laser/antenna_injection.py
```diff
@@ -289,272 +327,101 @@ def deposit( self, fld, fieldtype ):
...
-        # Add the values
-        if ir < Jr.shape[1]:
-            Jr[iz_min, ir] += Jr_buffer[m, 0, ir]
-            Jr[iz_min+1, ir] += Jr_buffer[m, 1, ir]
 
-            Jt[iz_min, ir] += Jt_buffer[m, 0, ir]
-            Jt[iz_min+1, ir] += Jt_buffer[m, 1, ir]
+        elif fieldtype == 'J' :
+            # Calculate the relativistic momenta from the velocities
+            # The gamma is set to 1 both here and in the deposition kernel
```
RemiLehe, line 411: Could you add a comment here saying that this works because the current deposition in fact only depends on `ux*inv_gamma`, `uy*inv_gamma` and `uz*inv_gamma`, so as long as we make sure that this product correspond to `vx`, `vy`, `vz`, we can indeed pass `inv_gamma = 1.`.

RemiLehe (approved):

RemiLehe (changes requested):
Thanks for this PR! I have two minor requests regarding comments/documentation.

lauridsj:
Sure! I totally forgot about the inv_gamma thing, that should of course be explained. I will do this today later.

lauridsj:
I added some comments - hope it is better now.

RemiLehe (approved):
Great! Thanks a lot for this PR! This makes the code much cleaner and more consistent!

PR #475, opened 2020-09-07
Fix scoping issues for external fields on the GPU

lauridsj:
This PR fixes a bug introduced in the latest overhaul of the external fields on GPU (#470) where scoping issues can lead to wrong results for multiple external fields. This was due to the external field GPU kernels being compiled not at the time of definition, but at the time of the first call.

To fix this, this PR adds an option to immediately specialize a GPU kernel with given function signature for the `compile_cupy` decorator. This option is then used for the external fields.


RemiLehe (approved):

RemiLehe:
Awesome! Thanks a lot! 🎉

PR #477, opened 2020-09-09
PICMI: read PSATD order in z from field solver

EZoni:
With this PR the order of the finite stencil along z for the Maxwell field solver is read directly from the member `stencil_order` (vector of integers) of the class `PICMI_ElectromagneticSolver`, instead of being hard-coded as `n_order=32`.

If `stencil_order` is not initialized in the solver, the default is set to `n_order=-1`, which corresponds to infinite order and matches the default in the `Simulation` class. Not sure which default order you would like to keep for the PICMI interface.


RemiLehe:
Great! Thanks a lot for this PR! ✨

PR #483, opened 2020-09-18
Revert antenna to previous behavior

RemiLehe:
This PR reverts the change from #461, because we recently observed that it unexpectedly changed the results of some recent boosted-frame simulations. 

@hightower8083 Sorry for the delay in incorporating this feature. I'll investigate the changes and will merge into `dev` again once this is cleared up.


hightower8083:
no problem, @RemiLehe ! I'm not sure there was a noticeable performance gain from this feature in my sims. 
Anyway in a case of a very heavy antenna, the same thing can be realised by doing restart-from-checkpoint after the laser is injected and with antenna removed.

PR #486, opened 2020-10-22
Improved performance of Flattened Gaussian laser

lauridsj:
This PR greatly improves the performance of the Flattened Gaussian laser profile. This is especially notable when used together with the laser antenna, where the laser profile calculation has before often dominated the runtime of a simulation when the order of the Flattened Gaussian is large (e.g. for N=100).

To achieve this, the Flattened Gaussian is implemented as a recursive sum instead of the sum over Laguerre-Gaussian modes it was before. This reduces the number of iterations needed from N! to N, where N is the order of the Flattened Gaussian.


RemiLehe:

Issue #488, opened 2020-11-12
Failed in nopython mode pipeline (step: convert to parfors)


hightower8083:
Doing a case on CPUs (MPI+openMP) on our lab cluster (_centos8_), I've stumbled across the following issue
```
raise ValueError("Cannot add edge as dest node %s not in nodes %s" %
ValueError: Failed in nopython mode pipeline (step: convert to parfors)
Cannot add edge as dest node 10 not in nodes {288, 68, 102, 326, 264, 42, 76, 334, 14, 92, 242, 24, 26, 347, 220}
```
which results from some issue with the nested `while` loop and can appear from numba methods either in [ionisation](https://github.com/fbpic/fbpic/blob/8530cd48fd5f5939e651143f7be37dadbafa6d5a/fbpic/particles/elementary_process/ionization/numba_methods.py#L54) or in cross-deposition ([here](https://github.com/fbpic/fbpic/blob/8530cd48fd5f5939e651143f7be37dadbafa6d5a/fbpic/fields/numba_methods.py#L100) and [here](https://github.com/fbpic/fbpic/blob/8530cd48fd5f5939e651143f7be37dadbafa6d5a/fbpic/fields/numba_methods.py#L263)). 

As mentioned in Numba [issue 5156](https://github.com/numba/numba/issues/5156) this problem is related to the `python>=3.8` -- it was also mentioned in #436 but was not relevant there. 

As was suggested by numba people, I've patched this bug in #489 with dummy `+=0` operations  after `while`-loops. This has fixed the issue for me, but feel free to proposed more thoughtful solutions (downgrading python is not always the best option)

PR #489, opened 2020-11-12
patch for the numba issue with nested while loop in `py>=3.8`

hightower8083:
This PR addresses the error
```
raise ValueError("Cannot add edge as dest node %s not in nodes %s" %
ValueError: Failed in nopython mode pipeline (step: convert to parfors)
Cannot add edge as dest node 10 not in nodes {288, 68, 102, 326, 264, 42, 76, 334, 14, 92, 242, 24, 26, 347, 220}
```
which results from some issue with the nested `while` loop and appears from numba methods with `py>=3.8`. More details are discussed in #488

### DIFF
--- a/fbpic/fields/numba_methods.py
+++ b/fbpic/fields/numba_methods.py
@@ -119,7 +119,7 @@ def numba_correct_currents_crossdeposition_standard( rho_prev, rho_next,
 
             # Increment ir
             ir += 1
-
+        ir += 0
     return
 
 @njit_parallel
@@ -285,6 +285,7 @@ def numba_correct_currents_crossdeposition_comoving(
 
             # Increment ir
             ir += 1
+        ir += 0
 
     return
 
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for reporting and fixing this issue!
Those `while` loops were introduced a long time ago, due to a quirk in an old version of `numba`.
But I think that this quirk has been fixed since a long time. As a result, I tried to convert those `while` loops to `for` loops in #490 . Could you test the code in this PR with python 3.8 and check whether fixes your issue?

hightower8083:
Yes, #490 works fine on my config with  `py@3.8.6` and `numba@0.51.2`, and i totally agree that dropping these while loops is the way ("this is the way" ©). So please go ahead and merge it unless you wanna test it with older numba vesrions. cheers!

RemiLehe:
Thanks for confirming! :)

PR #491, opened 2020-11-12
[WIP] Use latest version of numba and python

RemiLehe:
Currently, our CI uses an outdated version of numba. 
The reason for this is that the newest version of `numba` does not compile FBPIC (on CPU), as it seems to have issues with some of our `while` loops (esp. when parallelized).

This PR modifies the code to suppress some of these `while` loops. 

However, the current PR still fails, due to some bug in the ionization test.


fbpic/particles/elementary_process/compton/numba_methods.py
```diff
@@ -174,91 +174,91 @@ def scatter_photons_electrons_numba(
                     cos_phi = 1.
                     sin_phi = 0.
 
-            # Loop through the number of scatterings for this electron
-            for i_scat in range(nscatter_per_elec[i_elec]):
```
RemiLehe, line 178: All the changes below consist in reindenting this `for` loop, so that it is located *within* the `if` condition `if nscatter_per_elec[i_elec] > 0:`. It was found that this fixes some compilation errors.

PR #494, opened 2020-11-25
Add phase space diagnostics

tom48701:
### Ad hoc phase space diagnostics for fbpic
Just a temporary measure until there is an official implementation

The core of it is the `numpy.histogramdd` function, and the resulting dataset is appended to the usual hdf5 output files. The axis limits, names and quanaity desposited are saved as attributes of the dataset.
The dataset of the histogram is located within the diag file at

`data/[iteration]/phase spaces/[species]/[diagnostic name]`

with the attributes

`quantity` : deposited quantity
`axis1` : (min, max)
`axis1Name` : axis name

with additional axes numbered sequentially in the order they are passed.

Currently each instance of the `PhaseSpaceDiagnostic` class describes one diagnostic. There's no reason why it couldn't be set up to process multiple diagnostics from a single instance, and in fact I wrote it that way initially, but in practice it very quickly gets messy passing long lists of arguments defining all aspects of the required diagnostics. 
I found it easier to keep track of things just defining a new instance for each diagnostic, So for the sake of keeping things simple this is what I settled on. (Passing multiple species at once still works though)

To do:

Tidy up the metadata. Since it's adapted from an existing class for a different type of diagnostic, the metadata generated is lacking in some areas and redundant in others.

### DIFF
--- a/fbpic/openpmd_diag/phasespace_diag.py
+++ b/fbpic/openpmd_diag/phasespace_diag.py
@@ -1,0 +1,540 @@
+# Copyright 2016, FBPIC contributors
+# Authors: Remi Lehe, Manuel Kirchen
+# License: 3-Clause-BSD-LBNL
+
+"""
+This file defines the class PhaseSpaceDiagnostic
+"""
+import os
+import h5py
+import numpy as np
+from scipy import constants
+from .generic_diag import OpenPMDDiagnostic
+from .data_dict import macro_weighted_dict, weighting_power_dict
+
+class PhaseSpaceDiagnostic(OpenPMDDiagnostic) :
+    """
+    Class that defines a phase space diagnostic to be performed.
+    """
+
+    def __init__(self, period=None, species={}, comm=None,
+        name=None, phase_space=[], bins=[], edges=None, 
+        move_with_window=True, deposit='w', select=None, write_dir=None, 
+        iteration_min=0, iteration_max=np.inf, dt_period=None ) :
+        """
+        Initialize a phase space diagnostic.
+
+        Parameters
+        ----------
+        period : int, optional
+            The period of the diagnostic, in number of timesteps.
+            (i.e. the diagnostic is written whenever the number
+            of iterations is divisible by `period`). Specify either this or
+            `dt_period`.
+
+        dt_period : float (in seconds), optional
+            The period of the diagnostic, in physical time of the simulation.
+            Specify either this or `period`
+
+        species : a dictionary of :any:`Particles` objects
+            The object that is written (e.g. elec)
+            is assigned to the particle name of this species.
+            (e.g. {"electrons": elec })
+
+        comm : an fbpic BoundaryCommunicator object or None
+            If this is not None, the data is gathered by the communicator, and
+            guard cells are removed.
+            Otherwise, each rank writes its own data, including guard cells.
+            (Make sure to use different write_dir in this case.)
+        
+        name: str, optional
+            Specify a custom name for the diagnostic. If left as None, an
+            automatically generated name is used based on the phase space
+            and quantity deposited
+            
+        phase_space : a list of strings
+            Specify the phase space over which to bin particles, of the form
+            ['uz']        (bin along the z momentum)
+            ['z','gamma'] (bin along z position and the particle gamma)
+            
+        bins: a list of ints
+            A list of ints specifying the number of bins in each dimension.
+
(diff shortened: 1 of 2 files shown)


fbpic/openpmd_diag/phasespace_diag.py
```diff
@@ -0,0 +1,540 @@
+# Copyright 2016, FBPIC contributors
+# Authors: Remi Lehe, Manuel Kirchen
```
RemiLehe, line 2:
```suggestion
# Copyright 2020, FBPIC contributors
# Authors: Thomas Wilson, Remi Lehe, Manuel Kirchen
```

fbpic/openpmd_diag/phasespace_diag.py
```diff
@@ -0,0 +1,540 @@
...
+            deposit = self.get_dataset( species, self.deposit, select_array, 
+                                        n_rank, Ntot )            
+            weights *= deposit
+            # set the deposition metadata
+            dset.attrs['deposit'] = self.deposit 
+        else:
+            dset.attrs['deposit'] = 'weight'        
+  
+        # generate the histogram
+        hist, edges = np.histogramdd(sample, bins=self.bins,
```
RemiLehe, line 479: I would suggest to also put this call to `np.histogramdd` in the `if self.rank == 0:` condition below:
only the MPI rank that writes the results needs to perform this calculation.

fbpic/openpmd_diag/phasespace_diag.py
```diff
@@ -0,0 +1,540 @@
...
+        hist, edges = np.histogramdd(sample, bins=self.bins, 
+                                        density=False, weights=weights,
+                                        range=bin_edges)
+        
+        # Fill out the dataset and metadata for the axes
+        if self.rank==0:
+            dset[:] = hist
+            for i in range(self.Ndims):
+                dset.attrs['axis%i'%(i+1)] = [edges[i][0], edges[i][-1]]
+                dset.attrs['axis%iName'%(i+1)] = '%s'%self.phase_space[i]
```
RemiLehe, line 488: As mentioned above, I would suggest to replace the above meta-data by `axisLabel`, `gridSpacing`, `gridGlobalOffset`.

RemiLehe (commented):
Thanks a lot for contributing this to FBPIC!
The code is well written and easy to understand. I have a few comments and suggestions, mainly related to the metadata. (see the inline comments and suggestions).

In order to update the current pull request accordingly, you can implement the changes on your local computer and then push them to the same branch, with `git`.

Thanks again!

tom48701:
I have implemented the attribute changes, as you suggested, now the attributes for an N-dimensional dataset are:

`gridUnitsSI` an array of length N
`gridUnitDimension` an array of length 7*N (as per https://github.com/openPMD/openPMD-standard/pull/193)
`axisLabels` a list of length N containing the axis names
`gridSpacing` and `gridGlobalOffset` are analagous to those in a field diagnostic

the only attributes that don't seem to have analagous terms in other diagnostics are those describing the deposited quantity, so this information is contained in the

`depositedQuantityUnit`, `depositedQuantityUnitSI` and `depositedQuantityUnitDimension` attributes

these can be easily changed or removed if needed.

In addition to these changes, I have also added an additional feature.
There is a new class; `CustomQuantity`, which contains instructions to compute a custom particle quantity from the 'basic' particle quantities.

Currently it is just a simple collection of attributes describing the name of the quantity for reference purposes, a function describing how to compute the quantity, a list of neccessary arguments to that function and the dimensions of the computed quantity.

A list of instances of this class can be passed to the `PhaseSpaceDiagnostic`, and these will be taken into consideration when performing the diagnostic.

For instance, the radial position of a particle is given by sqrt(x^2 + y^2). A `CustomQuantity` class instance to compute this would be initialised as follows:

```
def r_func(args):
	return( np.sqrt(args[0]**2+args[1]**2) )

CustomQuantity('r', r_func, ('x','y'), [1.,0.,0.,0.,0.,0.,0.])
```

Passing this to the diagnostic would open up the 'r' quantity to be used as either an axis or a deposited quantity. So far I have tested this with angular momentum and the particle pitch angle and it seems both functional and useful.

comments:

In the attributes, only the short name of the quantity is deposited, i.e. 'w' instead of 'weighting'. Maybe this should be changed so the full names are written for clarity?
I'm still not sure exactly what is required and what can be omitted from the metadata, so that probably still needs work

PR #495, opened 2020-11-25
When using `verbose=2`, print node name whenever possible

RemiLehe:
When using `verbose=2`, the node name was not printed *for single-MPI simulations*.
Yet this information can be useful even for single-MPI simulation. This PR corrects this.


lauridsj (approved):
Seems good!

PR #498, opened 2020-12-03
Fixed external fields for multiple species

### LINKED ISSUE #497: External fields are not correctly applied for multiple species
I believe I have found a bug regarding the external fields. When the `species` option of the external fields is set to `None`, the fields are supposed to be applied to all species in the simulation. However, when multiple species are present, it seems that the fields are sometimes not applied at all, e.g. in this simulation of a plasma lens:

divergence_bug

This does not appear when only one particle species is present, or when the `species` argument is set explicitly. 

I believe that the issue appears when one species has no particles - the code of the external fields then returns and does not apply the expression to the other species. I'll try to fix this and submit a PR.

lauridsj:
This PR fixes #497 regarding the application of external fields for multiple species (see there). I tested it and in my case, the problem was gone.

### DIFF
--- a/fbpic/lpa_utils/external_fields.py
+++ b/fbpic/lpa_utils/external_fields.py
@@ -195,7 +195,7 @@ class ExternalField( object ):
                 # Only apply the field if there are macroparticles
                 # in this species
                 if species.Ntot <= 0:
-                    return
+                    continue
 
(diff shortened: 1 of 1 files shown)


RemiLehe (approved):
Thanks a lot for this fix!

PR #500, opened 2020-12-15
Add reference to latest article on finite-order Galilean

RemiLehe:
This adds the reference both in the README and in the documentation.


MKirchen:
Thanks for this!

PR #501, opened 2020-12-16
Update README.md

hightower8083:
my affiliation update


RemiLehe:
Thanks! Could you also update the file `docs/source/index.rst` in the same PR? That file is used in order to generate the online documentation.

hightower8083:
ah right, forgot about this one.. Done!

PR #507, opened 2021-01-26
Redefinition of Mirror class

delaossa:
It adds functionality by letting the user specify the field modes to be zeroed at the mirror region. 
By default, all modes are set to zero.
The syntax has slightly changed: the user now must give the z start and end positions of the mirror in the lab frame (before it was only the start position and the width of the mirror in number of cells of the boosted frame).


MKirchen:
Thanks @delaossa

PR #516, opened 2021-02-28
Deactivate Python 2.7 in automated tests

RemiLehe:
Python 2.7 is no longer officially maintained, and it seems that some FBPIC dependencies are now failing - when installed from Python 2.7 channels.

This PR deactivates Python 2 in the automated tests. From now on, only Python 3 will be tested.


MKirchen:
Thanks!

Issue #520, opened 2021-03-14
simulation is not exiting on `cupy.cuda.memory.OutOfMemoryError`


hightower8083:
following the discussion in Slack :

I've also my own little question about cuda out-of-memory situation. When FBPIC runs out of GPU memory, it's not exiting but keeps hanging in a running state. We've discussed this behaviour once and if I remember correctly in such case it transfers simulation to the host memory, which is practically equivalent to stop.
When running on cluster it results in some confusion -- if code eventually gets out-of-memory in a long scan, it keeps consuming computation hours until I notice it (by checking frozen out_logs) and shut it down manually, or it hits the time-limit.
Is there a simple way to prevent this migration to the host memory and just crash instead ?

Here is the  [error file](https://github.com/fbpic/fbpic/files/6136901/err_1439400.e.txt) from the job manager

RemiLehe:
Thanks a lot for reporting this issue!
Would you have time to install [this branch](https://github.com/fbpic/fbpic/pull/521) and re-try the above-mentioned simulation? In principle, this PR should fix your issue.

hightower8083:
I confirm that issue is now resolved -- it returns a proper out-of-memory message and quits the simulation.
Thanx a lot @RemiLehe !
closing the issue

PR #521, opened 2021-03-14
Catch cupy memory error

RemiLehe:
It seems that the code that catches memory error was not updated when we switched from the `numba.cuda` allocation to the `cupy` allocation.

I also added the decorator `catch_gpu_memory_error` around the sorting routine (`cupy` needs to allocate temporary arrays under the hood, when performing the Radix Sort).


RemiLehe:
This PR was successfully tested in #520.
Therefore, I will merge it here.

PR #522, opened 2021-03-17
Fix the particle boost to use a less numerically lossy algorithm

dpgrote:
The algorithm to do the boost in `BoostConverter.boost_particle_arrays` was subject to loss of numerical accuracy. The primary reason is its use of vz which can be very close to 1, causing a loss of precision. One result of the loss of precision is that the transverse momenta were coming out different than the input values, even though the transform should not affect them.

The algorithm it was replaced with applies the transform directly to the momenta. The vz is only calculated to advance the particle position. The code is quite a bit simpler and the direct copy of the transverse momenta is explicit.

Note that I use a copy to get new_ux and y to avoid any overwriting issues. If the ux and uy passed into this routine are not used for anything afterwards, the copy is not needed.


RemiLehe (approved):
Thanks for these changes!

PR #525, opened 2021-03-23
Force TPB values for V100 to apply for A100

### LINKED ISSUE #506: Optimization for A100
For P100 and V100, we use an optimal number of threads per block ; we could do the same for A100.

MKirchen:
Results in small performance increase.

Closes #506.


RemiLehe (approved):
Thanks!

PR #532, opened 2021-04-29
Fix cupy 9 compatibility

MKirchen:
Cupy 9 requires a helper GPU array (instead of a scalar) that is passed as an argument to the cuBLAS dgemm call during the Hankel transform.

- [x] Test with cupy 8 and 9

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -33,7 +33,7 @@ jobs:
       name: Install dependencies
       run: |
         conda install --yes "numba<0.49" scipy h5py mkl pytest pyflakes python=${{ matrix.python-version }};
-        conda install --yes -c conda-forge mpi4py;
+        conda install --yes -c conda-forge mpich mpi4py;
         pip install openPMD-viewer;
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for this PR! I just tested it with `cupy 9`, but for some reason this failed (`Fatal Python error: Segmentation fault`).

Switching to numpy arrays fixes it, although I do not entirely understand why:
```
                self.alpha_ary = np.array(self.alpha, dtype=self.d_in.dtype)
                self.alpha = self.alpha_ary.ctypes.data
                self.beta_ary = np.array(self.beta, dtype=self.d_in.dtype)
                self.beta = self.beta_ary.ctypes.data
```

lauridsj:
I checked the solution with numpy arrays on cupy 9, it works and is as fast as the current dev branch with cupy 8.6. So this seems good!

RemiLehe:
On my side, I ran our CI tests with both `cupy 8.6` and `cupy 9`, and everything passes. I think this is ready to be merged.

PR #537, opened 2021-05-17
Pass plasma extent to Particles and ContinuousInjector

syoffe:
This PR passes the user-defined plasma extent (`p_zmin` and `p_zmax`) on to the `Particles` and `ContinuousInjector` classes. The values passed currently have been modified to within the initial domain. This change allows the continuous injector to check whether the loading region is within the plasma extent **before** creating particles that are eventually ignored due to having zero weight. This can save a lot of time for short plasma species with many particles per cell.

### DIFF
--- a/fbpic/particles/injection/continuous_injection.py
+++ b/fbpic/particles/injection/continuous_injection.py
@@ -17,7 +17,7 @@ class ContinuousInjector( object ):
     """
 
     def __init__(self, Npz, zmin, zmax, dz_particles, Npr, rmin, rmax,
-                Nptheta, n, dens_func, ux_m, uy_m, uz_m, ux_th, uy_th, uz_th ):
+                Nptheta, n, dens_func, ux_m, uy_m, uz_m, ux_th, uy_th, uz_th, p_extent=None ):
         """
         Initialize continuous injection
 
@@ -29,6 +29,7 @@ class ContinuousInjector( object ):
         self.Npr = Npr
         self.rmin = rmin
         self.rmax = rmax
+        self.p_extent = p_extent
         self.Nptheta = Nptheta
         self.n = n
         self.dens_func = dens_func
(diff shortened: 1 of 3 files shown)


fbpic/particles/injection/continuous_injection.py
```diff
@@ -181,9 +182,9 @@ def dens_func(x, y, z):
 
         # Create new particle cells
         # Determine the positions between which new particles will be created
-        Npz = self.nz_inject
         zmax = self.z_end_plasma
         zmin = self.z_end_plasma - self.nz_inject*self.dz_particles
+        Npz = 0 if (self.p_extent and (zmax < self.p_extent[0] or zmin > self.p_extent[1])) else self.nz_inject
```
RemiLehe, line 187: Could you first add an `if` condition here, to check that `p_extent` is not `None`?

RemiLehe (changes requested):
Thanks for this PR!
It seems that some tests are not passing ; I suggest a change which might fix that (hopefully).

fbpic/particles/injection/continuous_injection.py
syoffe: Hi Remi, sure I can add that. I thought the first test failing (when `self.p_extent` is `None`) would prevent the subsequent tests being performed, which I did try locally without any issues but perhaps this depends on the `python` version.

syoffe:
Hi Remi,

I had to modify `tests/test_linear_wakefield.py` to increase `p_zmax`. Plasma was previously defined for `39e-6 < z < 41e-6`, but since `zmax = 40e-6` the plasma extent is only used for the initial loading in the full domain. The continuous injector doesn't check, and so continued loading never-ending plasma as the window moved. The above PR does now check the plasma extent, and so only loaded 2 microns of plasma then vacuum again, so the `assert` comparisons for the wakefield were bound to fail.

I have reverted to
```python
Npz = 0 if (self.p_extent is not None and len(self.p_extent)>1 \
            and (zmax < self.p_extent[0] or zmin > self.p_extent[1])) \
        else self.nz_inject
```
but I would be happy to split it into nested `if` statements if preferred.

RemiLehe:
Sorry for the very late review!
After looking at the code more closely, I think that this would require some small modifications in order to work properly in the boosted frame and/or with a moving plasma. 
I will modify the PR accordingly in the next few days.

PR #538, opened 2021-06-04
Fix warnings for Numba 0.53 and greater

lauridsj:
This PR fixes a bunch of warnings that appear in the output for Numba versions 0.53 and greater during kernel compilation on GPU, as well as incompatibilities with upcoming versions of numba. It also slightly refactors the cupy kernel launcher by moving the code that creates a cupy kernel into a seperate function to reduce redundancy.

I tested the PR on GPU with numba versions 0.50.1, 0.53.1, and the current dev version (0.54.0-dev).


RemiLehe (approved):
Awesome! Thanks a lot for rapidly fixing the issue!

PR #559, opened 2021-10-25
Add Laser Pulse with Custom Spectrum

rob-shalloo:
Adding functionality to allow a user defined spectral amplitude and phase to be used to describe the temporal profile of the pulse.


RemiLehe:
I just updated the PR to make it compatible with the latest changes in FBPIC (i.e. separation into longitudinal and transverse profile).
There are still a few things to complete (marked as TODO).

fbpic/lpa_utils/laser/longitudinal_laser_profiles.py
```diff
@@ -183,3 +184,165 @@ def squared_profile_integral(self):
...
+        -----------
+        z0: float (in meter)
+            The initial position of the centroid of the laser
+            (in the lab frame), defined as :math:`z_0` in the above formula.
+
+        spectrum_file: file path
+            The path to a csv file containing 3 columns (no headers).
+            The three columns should represent wavelength (in m), spectral
+            amplitude (arb. units) and spectral phase (in radians).
+            Use a "\t" tab as the deliminator in the file.
```
RemiLehe, line 212: We should probably discuss the sign of the spectral phase here.

pots007:
Hey guys! 
I think something like this is really cool and important - I've been working on something similar and have actually independently come up with pretty much the same solution. In my solution, there's a few things that I think should be merged here:
- the spectral intensity needs to scaled to omega-domain: dI/domega = dI/dlambda * lambda ^2  / (2 * pi * c) where dI/dlambda is spectral intensity in the wavelength domain
- I would add support for files with only 2 columns with some if statements after `np.loadtxt`, eg `if filetext.shape[1] < 3: phase = np.zeros(wavelentgh.shape)`
- I think this is a good chance to add an option to provide spectral phase expansion terms - a tuple of (GDD, TOD, FOD) for example, that can then be used to change the pulse shape/duration etc. Need to agree on the sign of spectral phase though, as you said!
- In the current state, propagation direction is not taken into account when calculating the pulse in `evaluate` method
- In the evaluate method, I think `profile = np.tile(timeProfile[:, None, None], (1, z.shape[1], z.shape[2]))` might be faster than the for loop?
- Using the intensity weighted mean wavelength can lead to trouble, as in `LaserProfile` it is checked whether `k0` of both spatial and temporal profile agree. It might be better to have the user specify a central wavelength nevertheless and a warning issues if the mean wavelength is too far off?

I'm not sure how to work on this PR though, can I check it out remotely and then push to this PR? Or do I need to create a PR in my fork based on this PR, which can then be merged with the current PR? Sorry, I'm still a bit of a novice in git :)

Cheers,
Kris

RemiLehe:
@pots007 Thanks a lot for the comments. I agree with them.
I will try to update and finalize this PR within the next two days. You can then take a look and see whether you agree with the changes.

RemiLehe:
I updated the PR and took into account @pots007's comments.

There are a few things that we should still discuss (probably with @pots007, @rob-shalloo, @MKirchen and @soerenjalas, probably after the Winter break):
- I agree that the check on `k0` makes it a bit cumbersome to use this new class. Here is how I bypassed this issue in the test associated with this PR: https://github.com/fbpic/fbpic/pull/559/files#diff-ad6c58a96a3de321bb65899aa8af859f3fdf7c3d8dd71edf8e8aa2eb174933bfR89-R91
However, it is not very convenient either.
- As @pots007 mentioned, we should probably discuss how to include GDD, TOD, FOD. A related problem is: what should we do if the user provides a phase with a linear slope as a function the wavelength (or, more precisely, as function of the wavevector). In principle, this corresponds to a time delay in real space, but this conflicts with the parameter `z0` which the user also passes. (More precisely, if there is a linear slope in phase, the pulse will not be centered exactly at `z0`, which might confuse the user.) One solution would be to systematically fit and subtract the linear slope in the phase...

RemiLehe:
Discussed off-line:
- The check that the wavelength are equal in the transverse and longitudinal profiles will be converted to a warning.
- We will add an option to subtract the linear slope in phase (will be turned off by default).
- We will add GDD, TOD, FOD as optional parameters.
- We will also add the optional to pass arrays, along with helper functions to create the array corresponding to e.g. a Gaussian pulse etc.

In addition: 
- We need a documentation page for this new feature.
- This would need to test this with boosted-frame and/or antenna.

fbpic/lpa_utils/laser/longitudinal_laser_profiles.py
```diff
@@ -183,3 +184,127 @@ def squared_profile_integral(self):
...
+
+        More specifically, the temporal characteristics of the pulse are
+        calculated numerically via the spectral intensity and phase which
+        are provided to the class as a path to a csv file containing the data.
+
+        More specifically, the electric field computed according to:
+        .. math::
+
+            E(z,t) \propto \int_{-\infty}^{\infty}\!\! dk \;k
+                \sqrt{\\frac{dI}{d\lambda}(2\pi/k)}
```
RemiLehe, line 206:
```suggestion
                \sqrt{I(2\pi/k)}
```

fbpic/lpa_utils/laser/longitudinal_laser_profiles.py
```diff
@@ -183,3 +184,127 @@ def squared_profile_integral(self):
...
+                \,e^{i\phi(2\pi/k)}\,e^{i k (z-z0 -ct)}
+
+        where :math:`\\frac{dI}{d\lambda}` and :math:`\phi` are the
```
RemiLehe, line 209:
```suggestion
        where :math:`I` and :math:`\phi` are the
```

fbpic/lpa_utils/laser/longitudinal_laser_profiles.py
```diff
@@ -183,3 +184,127 @@ def squared_profile_integral(self):
...
+        spectral intensity provided in the csv file (as a function of the
+        wavelength :math:`\lambda=2\pi/k`). (The fact that the integrand
+        is multiplied by :math:`k` in the above formula is because
+        the csv file provides :math:`\\frac{dI}{d\lambda}`, not
+        :math:`\\frac{dI}{dk}`.)
```
RemiLehe, line 214:
```suggestion
        the csv file provides the intensity distribution over wavelengths 
        :math:`\lambda`, instead of over wavevectors :math:`k`.)
```

RemiLehe (changes requested):

RemiLehe (approved):

PR #565, opened 2021-12-05
Added if to handle mpi4py 4.0

pots007:
The current mpi4py version check fails when using mpi4py 4.0. Adding another if statement seems to fix this.


RemiLehe (approved):

RemiLehe:
Thanks a lot for fixing this!

PR #566, opened 2021-12-10
Fixes for PICMI 0.0.18

RemiLehe:
The API recently slightly changed in PICMI 0.0.18. This changes FBPIC accordingly.


fbpic/picmi/simulation.py
```diff
@@ -250,11 +250,15 @@ def add_diagnostic(self, diagnostic):
             iteration_max = diagnostic.step_max
         # Register field diagnostic
         if type(diagnostic) == PICMI_FieldDiagnostic:
+            if diagnostic.data_list isi None:
```
RemiLehe, line 253:
```suggestion
            if diagnostic.data_list is None:
```

PR #569, opened 2022-02-03
Added missing default argument

pots007:
Added missing default argument; when initialising with dt_period would need to set period=None manually otherwise


RemiLehe (approved):

RemiLehe:
Great! Thanks for fixing this!

PR #579, opened 2022-04-12
Fixes for picmi

dpgrote:
This fixes several issues with the picmi interface.
 - Changes were needed for the grid since the single quantities like `nr` and `zmax` are no longer saved as attributes
 - Allowed `None` to be used for the bounds of the plasma
 - Added diagnostics for names like `E` and `B`
 - Added fbpic specific arguments for the Simulation class
 - Added a change to the main loop so that the diagnostics are written out for the last time step (is there a better way to do this?)

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -481,10 +481,6 @@ class Simulation(object):
                 # were smoothed/corrected, and copy the data from the GPU.)
                 diag.write( self.iteration )
 
-            if i_step == N:
-                # Step "N+1" is started in order to get the final diagnostics written out
-                break
-
             # Push the particles' positions and velocities to t = (n+1/2) dt
             if move_momenta:
                 for species in ptcl:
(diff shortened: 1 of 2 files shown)


fbpic/main.py
```diff
@@ -418,7 +418,7 @@ def step(self, N=1, correct_currents=True,
         # -----------------------------
 
         # Loop over timesteps
-        for i_step in range(N):
+        for i_step in range(N+1):
```
RemiLehe, line 421:
```suggestion
        for i_step in range(N):
```

RemiLehe, line 421:
```suggestion
        for i_step in range(N):
```

fbpic/main.py
```diff
@@ -481,6 +481,10 @@ def step(self, N=1, correct_currents=True,
                 # were smoothed/corrected, and copy the data from the GPU.)
                 diag.write( self.iteration )
 
+            if i_step == N:
+                # Step "N+1" is started in order to get the final diagnostics written out
+                break
+
```
RemiLehe, line 487:
```suggestion

```

PR #592, opened 2022-05-22
Limit snapshot creation to time interval in back-transformed diagnostics

MKirchen:
This PR adds the possibility to limit the creation of snapshots to a specific time interval (in the lab frame) when using the back-transformed diagnostics in Lorentz-boosted frame simulations. 

The time interval is defined by two new parameters `t_min_snapshots_lab` and `t_max_snapshots_lab`. The parameters are similar to `iteration_min` and `iteration_max` in the standard OpenPMD diagnostics.

Note that the functionality of this PR can also be mimicked by deleting snapshots from the list of snapshots after creating them. For example:
```
# Create the diagnostics
sim.diags = [ BackTransformedFieldDiagnostic( ... ), ... ]
# Delete the first 3 snapshots (from the first diagnostic)
del sim.diags[0].snapshots[:3]
```

Therefore, it is up to debate if this PR is really necessary.


RemiLehe (approved):

RemiLehe:
Great, thanks!

PR #593, opened 2022-05-22
Use add_laser_pulse in example scripts

MKirchen:
This PR updated the example scripts to use the newer laser API. First, a laser profile is created, e.g. through `GaussianLaser()`, and it is then added to the simulation by calling the function `add_laser_pulse(sim, laser_profile, ...)`. The boosted frame example now uses the `antenna` method to add the laser pulse.


RemiLehe (approved):

PR #604, opened 2022-08-03
Implement user-defined injection scheme

kli-jfp:
* The injection scheme works only with moving window velocity = 0
* Injection dictionary with key values 'p': injection period
	and 't': injection duration have been added to
	simulation class


kli-jfp:
This code uses the moving window with v=0 to inject particles in intervals of injection period (multiple of exchange period) for a user-defined injection duration. This is a quick way of injecting particles into the simulation with minimal changes to the code. If you have comments on how to do this in another (more efficient) way then please let me know.

Issue #602, opened 2022-07-28
ptx module error with cupy


kli-jfp:
I tried installing other programs earlier and managed to break something. Perhaps someone here has run into a similar issue.In fbpic/utils/cuda.py...

    module = cupy.cuda.function.Module()
    if numba_version[1] >= 53:
      ptx = next(iter(numba_kernel.overloads.values())).ptx
      module.load(bytes(ptx, 'UTF-8'))
    else:
      module.load(bytes(numba_kernel.ptx, 'UTF-8'))

When I run my python input script I get this error

    AttributeError: '_Kernel' object has no attribute 'ptx'

I tried reinstalling fbpic, cupy, numba, scipy, numba-scipy... to no use. (edited) 

cupy.show_config()
OS                           : Linux-5.15.0-41-generic-x86_64-with-glibc2.35
Python Version               : 3.10.4
CuPy Version                 : 11.0.0
CuPy Platform                : NVIDIA CUDA
NumPy Version                : 1.22.4
SciPy Version                : 1.7.3
Cython Build Version         : 0.29.28
Cython Runtime Version       : 0.29.28
CUDA Root                    : /usr/local/cuda-11.7
nvcc PATH                    : /usr/local/cuda-11.7/bin/nvcc
CUDA Build Version           : 11070
CUDA Driver Version          : 11070
CUDA Runtime Version         : 11070
cuBLAS Version               : (available)
cuFFT Version                : 10702
cuRAND Version               : 10210
cuSOLVER Version             : (11, 3, 5)
cuSPARSE Version             : (available)
NVRTC Version                : (11, 7)
Thrust Version               : 101500
CUB Build Version            : 101500
Jitify Build Version         : <unknown>
cuDNN Build Version          : None
cuDNN Version                : None
NCCL Build Version           : None
NCCL Runtime Version         : None
cuTENSOR Version             : None
cuSPARSELt Build Version     : None
Device 0 Name                : NVIDIA GeForce RTX 3050 Ti Laptop GPU
Device 0 Compute Capability  : 86
Device 0 PCI Bus ID          : 0000:01:00.0

I installed these earlier

    sudo apt install build-essential ccache cmake g++ git libfftw3-mpi-dev libfftw3-dev libhdf5-openmpi-dev libopenmpi-dev pkg-    config python3 python3-matplotlib python3-numpy python3-pandas python3-pip python3-scipy python3-venv

Can this be creating some conflicts maybe?

kli-jfp:
[...]

PR #606, opened 2022-08-03
Make FBPIC compatible with numba version 0.56: Remove use of `ptx`

RemiLehe:
This PR addresses #602 attempts to remove the use of `ptx`, which is not supported anymore in numba `0.56`.
The fix is based on this line of numba code:
https://github.com/numba/numba/pull/8191/files#diff-ad6746f9263e40f10fa2a85537c0da56a1a8abfd7a0743ec0c06e3a14e3df0e7L193
Importantly, this change is in an `if` condition, based on the detected version of `numba`, so that the code remains compatible with previous versions of `numba`.

**Current status**:
Update: the issue described below was fixed in `numba 0.56.2`, thanks to [this PR](https://github.com/numba/numba/pull/8310>). As a result, FBPIC does not work with `numba 0.56.0` on GPU, but works with both previous versions of `numba` and later versions of `numba`.
~~The fix works with `numba 0.55.2` (for which it is however not required, since 0.55.2 still supports `.ptx`).~~
~~However, it fails for `numba 0.56.0` due to another error:~~

```
Failed in cuda mode pipeline (step: nopython frontend)
No implementation of function Function(<function ol_array_real at 0x7fd22405bc70>) found for signature:
 
 >>> ol_array_real(array(complex128, 2d, C))
 
There are 2 candidate implementations:
      - Of which 2 did not match due to:
      Overload in function 'ol_array_real': File: numba/np/arrayobj.py: Line 2620.
        With argument(s): '(array(complex128, 2d, C))':
       Rejected as the implementation raised a specific error:
         NumbaTypeError: Failed in nopython mode pipeline (step: nopython frontend)
       Only accept returning of array passed into the function as argument
  raised from /home/rlehe/miniconda3/envs/fbpic-new/lib/python3.10/site-packages/numba/core/typed_passes.py:150

During: typing of get attribute at /home/rlehe/Documents/codes/fbpic/fbpic/particles/deposition/cuda_methods.py (182)

File "../../../fbpic/particles/deposition/cuda_methods.py", line 182:
def deposit_rho_gpu_linear(x, y, z, w, q,
    <source elided>
            # Mode 0
            cuda.atomic.add(rho_m0.real, (iz0, ir0), R_m0_00.real)
            ^
```

### DIFF
--- a/fbpic/utils/cuda.py
+++ b/fbpic/utils/cuda.py
@@ -386,7 +386,8 @@ if cuda_installed:
             # numba kernel
             module = cupy.cuda.function.Module()
             if numba_version[1] >= 53:
-                ptx = next(iter(numba_kernel.overloads.values())).ptx
+                kernel = next(iter(numba_kernel.overloads.values()))
+                ptx = kernel._codelibrary.get_asm_str()
                 module.load(bytes(ptx, 'UTF-8'))
             else:
                 module.load(bytes(numba_kernel.ptx, 'UTF-8'))
--- a/fbpic/utils/cuda.py
+++ b/fbpic/utils/cuda.py
@@ -385,9 +385,12 @@ if cuda_installed:
             # Create a Cupy kernel module and load the PTX code of the
             # numba kernel
             module = cupy.cuda.function.Module()
-            if numba_version[1] >= 53:
+            if numba_version[1] >= 56:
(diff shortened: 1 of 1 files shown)


RemiLehe:
It seems that there is an issue with atomic adds to complex arrays in numba 0.56.
I reported the issue here:
https://github.com/numba/numba/issues/8309

PR #612, opened 2022-08-24
[WIP] Implement collision scheme in FBPIC

kli-jfp:
The collision scheme:
	1. Particles are grouped into random pairs
	2. The particle momenta are transformed to the
	Center of Mass (COM) frame
	3. Calculate coulomb logarithm and deflection angle
	4. Perform collisions in COM and transform back to
	lab frame
* This work implements the collision scheme proposed
	in Phys. Plasmas 19, 083104 (2012).
* It also implements the improvements proposed
	by the Smilei team, i.e. the deflection angle
	has been modified for better accuracy.

How to implement in input deck:

sim.collisions = [
        MCCollisions(elec, prot, use_cuda, coulomb_log = coulomb_log)
    ]

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -468,7 +468,7 @@ class Simulation(object):
             for collision in self.collisions:
                 if self.iteration % collision.period == 0 \
                     and self.iteration >= collision.start:
-                    collision.handle_collisions( fld, 0.5*dt )
+                    collision.handle_collisions( fld, dt )
             # Cell quantities need to be recalculated
             for species in ptcl:
                 species.calc = False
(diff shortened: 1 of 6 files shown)


kli-jfp:
I noticed that I have mistaken the normalized momentum u with the velocity in handle_collisions. I will fix this.

kli-jfp:
Woops, it's seems my original understanding was correct. The normalized momenta is indeed a velocity. The confusion came from reviewing Smilei's code where they call it "momentum" and "p". I will normalize all the equations in the collision code to make sure it is correct.

PR #617, opened 2022-09-29
Make the install of MKL more flexible for Windows

eurazov:
MKL numbers libraries now. Anaconda seems to have them numbered with '1'. Implemented in the same manner as it is done for Linux.


RemiLehe:
Thanks a lot for this fix!

PR #630, opened 2023-01-05
Added particle tracer option

pots007:
Hi, 

I added a few tiny modifications to be able to create particle species as tracers, ie particles that move as particles with the given mass and charge should, but contribute no current. This allows these to be used as passive probes of the plasma.

The attached video shows what this does: when creating a density bump in a plasma and shooting a high intensity laser through it, the background plasma is not affected at all if the species is defined as a tracer particle. If the bump is made of normal electrons, the background plasma is affected by the extra injected charge. I also have a script I used for this, but can't seem to attach it here...

Cheers,
Kris

https://user-images.githubusercontent.com/4076141/210784309-99c3e69d-66c1-40d2-bfe0-ea88f08dd362.mp4

### DIFF
--- a/fbpic/main.py
+++ b/fbpic/main.py
@@ -614,9 +614,10 @@ class Simulation(object):
         """
         # Shortcut
         fld = self.fld
-        # If no species_list is provided, all species and antennas deposit
+        # If no species_list is provided, all non-zero-current species
+        # and antennas deposit
         if species_list is None:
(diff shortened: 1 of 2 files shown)


RemiLehe:
Thanks for this PR. The automated tests fail because of a recent change in PICMI. This should be fixed once #631 is merged.

RemiLehe (approved):

PR #645, opened 2023-06-12
Support for lasy files

RemiLehe:
This PR adds support to read `lasy` files in FBPIC.

### DIFF
--- a/fbpic/lpa_utils/laser/laser_profiles.py
+++ b/fbpic/lpa_utils/laser/laser_profiles.py
@@ -856,7 +856,6 @@ class FromLasyFileLaser( LaserProfile ):
             emitted. This can be used in order to introduce a time delay that was
             not originally present in the ``lasy`` file.
 
-
         .. warning::
 
             This laser profile can only be emitted with the ``antenna`` method
@@ -869,15 +868,38 @@ class FromLasyFileLaser( LaserProfile ):
         # Open and read the lasy file
         f = h5py.File( filename, mode="r" )
         dset = f['/data/0/meshes/laserEnvelope']
+        self.omega = dset.attrs['angularFrequency']
+        self.pol = dset.attrs['polarization']
+        self.t_min_lasy = dset.attrs['gridGlobalOffset'][0]
+
+        # Define interpolation function depending on the geometry
+        if dset.attrs['geometry'].decode() == 'thetaMode':
+            self.define_thetaMode_interp_function(dset)
+        elif dset.attrs['geometry'].decode()  == 'cartesian':
+            self.define_cartesian_interp_function(dset)
+        else:
+            raise RuntimeError(
+                "Unknown geometry for lasy file %s: %s" \
+                %(filename, dset.attrs['geometry'])
+            )
+
+
+    def define_thetaMode_interp_function(self, dset):
+        """
+        Set the attribute `interp_function`, in the case case
+        where the lasy file has `thetaMode` geometry
+
+        Parameter:
+        ----------
+        dset: an h5py dataset object
+            The object that contains the lasy envelope data
+        """
         env_data = np.array(dset)
         dt_data, dr_data = dset.attrs['gridSpacing']*dset.attrs['gridUnitSI']
         inv_dt_data = 1./dt_data
         inv_dr_data = 1./dr_data
         _, nt, nr = env_data.shape
         n_modes = int(env_data.shape[0]/2) + 1
-        self.omega = dset.attrs['angularFrequency']
-        self.pol = dset.attrs['polarization']
-        self.t_min_lasy = dset.attrs['gridGlobalOffset'][0]
 
         @numba.vectorize
         def interp_function(x, y, t):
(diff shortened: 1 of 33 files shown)


fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -926,7 +925,7 @@ def E_field( self, x, y, z, t ):
         env = self.interp_function(x, y, (t-self.t_start) )
         # Add laser oscillations
         E = (env * np.exp(
-            -1.j*self.omega * (t - self.t_start + self.t_min_lasy)
```
hightower8083: hey just to remind @RemiLehe, that i'm not sure I agree with that -- this  would mess up the phase that in lasy is defined on its own time axis

RemiLehe: Ah, yes. Do you mean that it should simply be:
```
(t - self.t_start)
```

hightower8083: oh, I think my comment outdated -- there was a change with `self.t_min_lasy` being removed, but now it's reverted. false alarm =)

fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -835,3 +837,175 @@ def E_field( self, x, y, z, t ):
...
+
+        Parameters
+        ----------
+        filename: string
+            The path to the ``lasy`` file.
+
+        t_start: float (in seconds), optional
+            Physical time (in the simulation), at which the laser will start being
+            emitted. This can be used in order to introduce a time delay that was
+            not originally present in the ``lasy`` file.
```
RemiLehe, line 857:
```suggestion
            Physical time (in the simulation), at which the laser will start being
            emitted. This can be used in order to introduce a time delay that was
            not originally present in the ``lasy`` file. Note that the beginning of 
            the time axis in the ``lasy`` file is forced to be zero, when initializing
            the laser with FBPIC (irrespective of the metadata for `tmin` that is 
            actually present in the ``lasy`` file.)
```

fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -835,3 +837,175 @@ def E_field( self, x, y, z, t ):
...
 
         return( Ex.real, Ey.real )
+
+class FromLasyFileLaser( LaserProfile ):
+    """Class that emits a laser from a lasy file"""
+
+    def __init__(self, filename, t_start=0.):
+        """
+        Define a laser whose profile is determined by a
+        `lasy <https://lasydoc.readthedocs.io/en/latest/>`_ file.
```
RemiLehe, line 847:
```suggestion
        `lasy <https://lasydoc.readthedocs.io/en/latest/>`_ file.
        
        When the laser is initialized by this function, FBPIC forces the beginning of 
        the time axis in the ``lasy`` file to be zero (irrespective of the metadata for 
        ``tmin`` that is actually present in the ``lasy`` file). This convention was chosen
        for convenience, because ``tmin`` in ``lasy`` could otherwise result in large delays
        before emitting the laser, especially when using ``lasy``'s ``propagate`` feature.
```

fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -835,3 +837,175 @@ def E_field( self, x, y, z, t ):
...
+
+        t_start: float (in seconds), optional
+            Physical time (in the simulation), at which the laser will start being
+            emitted. This can be used in order to introduce a time delay that was
+            not originally present in the ``lasy`` file.
+
+        .. warning::
+
+            This laser profile can only be emitted with the ``antenna`` method
+            (not with the ``direct`` method).
```
RemiLehe, line 869:
```suggestion
            (not with the ``direct`` method).
            
    Example
    ---------
    .. code-block:: python
        # Creating the lasy file
        laser_profile = GaussianProfile(wavelength,polarization,
                                   energy,spot_size,pulse_duration,t_peak=0)
        dimensions     = 'rt'                              # Use cylindrical geometry
        lo             = (0,-2.5*pulse_duration)           # Lower bounds of the simulation box
        hi             = (5*spot_size,2.5*pulse_duration)  # Upper bounds of the simulation box
        num_points     = (300,500)                         # Number of points in each dimension
        laser = Laser(dimensions,lo,hi,num_points,laser_profile)
        laser.propagate(-1e-3)  # Propagate backwards by 1 mm
        laser.write_to_file('lasy_laser', 'h5')
        # Note that, in the lasy file, tmin is now a large, negative number.
        # (The peak of laser intensity still occurs 2.5*pulse_duration after tmin.)
        # FBPIC ignores tmin when reading the file, and sets the start of 
        # the lasy time axis to zero instead. So, by default, the peak of 
        # the laser intensity would occur at `t= 2.5*pulse_duration` in FBPIC.
        laser_profile = FromLasyFileLaser( 'lasy_laser_00000.h5', t_start=0.5*pulse_duration )
        add_laser_pulse(sim, laser_profile, method='antenna', z0_antenna=0)
        # Here, modify the `t_start`, which will result in the peak of intensity being 
        # emitted at `(2.5+0.5)*pulse_duration` instead of `2.5*pulse_duration`.
        # Since the `z0_antenna` was set to `0` here, this is as if the centroid 
        # of the laser would have been initialized at `z = -3*c*pulse_duration` at `t=0`
        # (with then ``direct`` method).
```

fbpic/lpa_utils/laser/laser_profiles.py
RemiLehe:
```suggestion
            not originally present in the ``lasy`` file. (As explained above, FBPIC ignores any 
            initial time offset in the ``lasy``. This offset is replaced by `t_start` (or zero if unspecified).
```

fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -835,3 +837,175 @@ def E_field( self, x, y, z, t ):
...
+        """
+        Define a laser whose profile is determined by a
+        `lasy <https://lasydoc.readthedocs.io/en/latest/>`_ file.
+
+        Parameters
+        ----------
+        filename: string
+            The path to the ``lasy`` file.
+
+        t_start: float (in seconds), optional
```
RemiLehe:
```suggestion
        t_start: float (in seconds), optional, default: 0
```

fbpic/lpa_utils/laser/laser_profiles.py
```diff
@@ -835,3 +837,175 @@ def E_field( self, x, y, z, t ):
...
+
+    def E_field( self, x, y, z, t ):
+        """
+        See the docstring of LaserProfile.E_field
+        """
+        # Perform interpolation from envelope data
+        env = self.interp_function(x, y, (t-self.t_start) )
+        # Add laser oscillations
+        E = (env * np.exp(
+            -1.j*self.omega * (t - self.t_start + self.t_min_lasy)
```
RemiLehe, line 1041:
```suggestion
            -1.j*self.omega * (t - self.t_start + self.t_min_lasy)
            # t_min_lasy is used here in order to have the same CEP as the
            # one that was meant when creating the lasy file
```

MKirchen (approved):
Thanks for this PR

PR #646, opened 2023-06-13
Update format and dependencies of CI test

RemiLehe:
This updates the CI tests so that it:
- Uses a recent version of Python
- Uses the latest version of `numba`
- Avoids using the deprecated command `python setup.py test`

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -6,7 +6,7 @@ jobs:
   test:
     strategy:
       matrix:
-        python-version: [3.6]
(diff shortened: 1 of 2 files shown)


.github/workflows/unix.yml
```diff
@@ -6,7 +6,7 @@ jobs:
   test:
     strategy:
       matrix:
-        python-version: [3.6]
+        python-version: [3.7]
```
RemiLehe:
```suggestion
        python-version: [3.11]
```

PR #654, opened 2023-07-24
Avoid warning and use new interface in `test_boosted_particle_output.py`

RemiLehe:
The file `test_boosted_particle_output.py` used to raise a warning about 0 charge, during the space-charge calculation. I fixed the warning by deactivating space-charge calculation. 
In addition, I used the new function `add_particle_bunch_gaussian` instead of the old function `add_elec_bunch_gaussian`.


MKirchen (approved):
Thanks 🙏

PR #656, opened 2023-08-02
Correct positions in antenna emission

RemiLehe:
In the function that updates the velocity of the antenna macroparticles, there was a mismatch between the position and time being used. More specifically, when calling this function, the positions are known at **integer** time, while the time `t` being passed is the one at **half-integer** time. In theory, this mismatch has important consequences when running in the boosted frame, since in this case, we perform a Lorentz transform that mixes position and time.

It seems that this change has little effect on existing tests ; maybe this is because all the laser profiles that we test are a function `z-ct`. However, I did notice that this PR has a strong effect on the profiles from lasy file: https://github.com/fbpic/fbpic/pull/645


MKirchen (approved):
Thanks for fixing this bug!

PR #657, opened 2023-08-02
Remove obsolete "python setup.py test" command

RemiLehe:
The command `python setup.py test` is obsolete and should be replaced by `python -m pytest tests --ignore=tests/unautomated` (See https://github.com/fbpic/fbpic/pull/646).

This PR updates various comments and docstrings, to remove the obsolete command.


MKirchen (approved):
Thanks

PR #661, opened 2023-08-28
Fix version of numexpr to avoid error message in CI

RemiLehe:
It seems that `numexpr 2.8.5` has some breaking changes that result in errors when certain floating point notations. For instance:
```
numexpr.evaluate('1.e24')
```
results in 
```
Expression 1.e24 has forbidden control characters.
```

For now, I will fix the version of `numexpr` to avoid this issue.

### DIFF
--- a/.github/workflows/unix.yml
+++ b/.github/workflows/unix.yml
@@ -50,7 +50,7 @@ jobs:
     - shell: bash -eo pipefail -l {0}
       name: PICMI test
       run: |
-        pip install wget picmistandard numexpr periodictable
+        pip install wget picmistandard numexpr<2.8.5 periodictable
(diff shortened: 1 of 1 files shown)


MKirchen (approved):

.github/workflows/unix.yml
```diff
@@ -50,7 +50,7 @@ jobs:
     - shell: bash -eo pipefail -l {0}
       name: PICMI test
       run: |
-        pip install wget picmistandard numexpr periodictable
+        pip install wget picmistandard numexpr<2.8.5 periodictable
```
RemiLehe:
```suggestion
        pip install wget picmistandard numexpr==2.8.4 periodictable
```

PR #665, opened 2023-10-10
Check the lasy version when reading lasy files

RemiLehe:
Since the format of `lasy` files has been evolving, we need to check the version number of the file, before attempting to read it.


MKirchen (approved):
Thanks for the PR

PR #672, opened 2023-12-01
Spin tracking

pots007:
This PR implements spin tracking in FBPIC. 

The API is (hopefully) straightforward: the `Particle` class now has a method `activate_spin_tracking`, which turns on spin tracking by creating a new instances on `SpinTracker`. All necessary arrays are stored inside this object. The user can set the average polarisation and have the spins be generated randomly or set to a fixed value.


tests/test_spin_ionization.py
```diff
@@ -0,0 +1,132 @@
...
+dt = (zmax - zmin) / Nz / c   # Timestep (seconds)
+
+
+def dens_func( z, r ) :
+    """Returns relative density at position z and r"""
+    # Allocate relative density
+    n = np.ones_like(z)
+    #n = np.where(r < rmax/Nr, n, 0.)
+    #n = np.where(np.abs(z) < zmax/Nz, n, 0.)
+    # n = np.where( z<ramp_start+ramp_length, (z-ramp_start)/ramp_length, n )
```
RemiLehe:
```suggestion
```

RemiLehe: Was it your intention to have a uniform density in this test (i.e. no ramps)?
If yes, you actually don't need to pass a `dens_func`.

tests/test_spin_ionization.py
```diff
@@ -0,0 +1,132 @@
+"""
+Usage :
+from the top-level directory of FBPIC run
+$ python spin_planewave_test.py
```
RemiLehe:
```suggestion
$ python spin_ionization.py
```

tests/test_spin_ionization.py
```diff
@@ -0,0 +1,132 @@
...
+    matplotlib.use('Agg')
+    import matplotlib.pyplot as plt
+    fig = plt.figure()
+
+    # ax = fig.add_subplot(projection='3d')
+    ax = fig.add_subplot(111, projection='3d')
+    ax.set_xlabel('sx')
+    ax.set_ylabel('sy')
+    ax.set_zlabel('sz')
+
```
RemiLehe:
```suggestion
```
It does not look like you are plotting anything here.

tests/test_spin_ionization.py
```diff
@@ -0,0 +1,132 @@
...
+    elec_total = 5 * atoms_Cl.Ntot
+
+    # run until we have enough ionized electrons...
+    elec_sum = 0
+    step = 0
```
RemiLehe:
```suggestion
```

tests/test_spin_ionization.py
```diff
@@ -0,0 +1,132 @@
...
+        # step and collect the data
+        sim.step(1, show_progress=False)
+        print('Step  Level  Ntot  N(sz==1)')
+        for level, elec in elec_dict.items():
+            elec_sum += elec.Ntot
+            if level < 6:
+                print(f'{step:4d}  {level+1:5d}  {elec.Ntot:4d} ' +
+                      f'{np.sum(elec.spin_tracker.sz==1.):8d}')
+        print('*'*40)
+
```
RemiLehe: It is good to have `print` statements, but eventually, the function should also automatically check the quantities of interest, so that there is no need for a human to look at the test results every time the tests run.

Here this could be (outside of the `while` loop, at the end):
```
if level < 1:
    assert elec.spin_tracker.sz.mean() > 0.99
else:
    assert abs(elec.spin_tracker.sz.mean()) < 0.1

File: tests/test_spin_planewave.py [hunk 6]
```diff
@@ -0,0 +1,293 @@
...
+
+(see e.g. Salamin & Faisal. Phys. Rev. A 54 (5 Nov. 1996), pp. 4383–4395.
+DOI : 10.1103/PhysRevA.54.4383)
+
+For these equations of motion, one can solve the Bargmann-Michel-Telegdi
+equation to obtain the following solutions:
+sx(phi) = sx_i*cos(I[phi]) - sz_i*sin(I[phi])
+sy(phi) = sy_i
+sz(phi) = sz_i*cos(I[phi]) + sx_i*sin(I[phi])
+I(phi) = a0*anom*sin(phi) + 2*arctan(a0*sin(phi)/(gamma_i - uz_i))
```
RemiLehe, line 41: Is the above solution for the spin evolution included in the Salamin & Faisal? If no, could you add a reference that gives the above solution for the spin evolution?

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
         if hasattr( self, 'int_sorting_buffer' ) is False and self.use_cuda:
             self.int_sorting_buffer = np.empty( self.Ntot, dtype=np.uint64 )
 
+    def activate_spin_tracking(self, sx_m=0., sy_m=0., sz_m=1.,
```
RemiLehe, line 514: Could you add this in the API documentation by adding:
```
.. automethod:: activate_spin_tracking
```
in this file:
https://github.com/fbpic/fbpic/blob/dev/docs/source/api_reference/particles.rst

tests/test_spin_ionization.py
RemiLehe: When I tried to run this test on GPU, I got the following error:
```
            # Split the particle array into the 3 buffers on the GPU
>           particle_array = getattr( attr_list[i_attr][0], attr_list[i_attr][1] )
E           IndexError: list index out of range
```
Can you reproduce this error on your side? If yes, would you be able to fix it?

tests/test_spin_planewave.py
```diff
@@ -0,0 +1,293 @@
...
+        add_gridlines()
+        plt.legend(['Analytic', 'FB-PIC'])
+        plt.show()
+
+        # save data as uncompressed .npz file
+        #np.savez(dir+'data.npz', t=t/T, t_mhalf=(t-dt/2)/T, ux=ux,
+        #         ux_analytic=ux_analytic, uz=uz, uz_analytic=uz_analytic, sx=sx,
+        #         sx_analytic=sx_analytic, sz=sz, sz_analytic=sz_analytic,
+        #         x=k0*(x+lambda0), z=k0*(z+lambda0))
+
```
RemiLehe:
```suggestion
```

tests/test_spin_planewave.py
```diff
@@ -0,0 +1,293 @@
...
+        rtol = 1e-3
+        # Interpolate to shift the analytic result by half time-step
+        ux_anal_shifted = np.interp(t[1:]-dt*0.5, t, ux_analytic)
+        assert np.allclose( ux_anal_shifted, ux[1:], atol=atol, rtol=rtol )
+        print(f'The momentum ux agrees with the theory to atol={atol},\n' +
+              'over the whole simulation box.')
+        sx_anal_shifted = np.interp(t[1:] - dt * 0.5, t, sx_analytic)
+        assert np.allclose(sx_anal_shifted, sx[1:], atol=atol, rtol=rtol)
+        print(f'The spin sx agrees with the theory to atol={atol},\n' +
+              'over the whole simulation box.')
```
RemiLehe: We only check one electron, so did you mean "over the duration of the simulation"?
```suggestion
        print(f'The momentum ux agrees with the theory to atol={atol},\n' +
              'over the whole duration of the simulation.')
        sx_anal_shifted = np.interp(t[1:] - dt * 0.5, t, sx_analytic)
        assert np.allclose(sx_anal_shifted, sx[1:], atol=atol, rtol=rtol)
        print(f'The spin sx agrees with the theory to atol={atol},\n' +
              'over the whole duration of the simulation.')
```

tests/test_spin_planewave.py
```diff
@@ -0,0 +1,293 @@
...
+        I = anom*a0sinphi + 2 * np.arctan(a0sinphi / (1+delta))
+        sz_analytic[i] = sz_i * np.cos(I) + sx_i * np.sin(I)
+        sx_analytic[i] = sx_i * np.cos(I) - sz_i * np.sin(I)
+
+    # Show the results
+    if show is False:
+        # Automatically check that the fields agree,
+        # to an absolute tolerance
+        atol = 0.01
+        rtol = 1e-3
```
RemiLehe, line 219: Out of curiosity, is this precision enough to check the effect of the anomalous momentum?

For instance, if you initialize the electrons with `anom=0` (i.e. `elec.activate_spin_tracking(sx_m=sx_i, sy_m=sy_i, sz_m=sz_i, anom=0)`) but compute the analytical results with the correct value of `anom`, does the test fail?

tests/test_spin_restart.py
```diff
@@ -0,0 +1,233 @@
+# Copyright 2017, FBPIC contributors
+# Authors: Kristjan Poder
+# License: 3-Clause-BSD-LBNL
+"""
+This test file is part of FB-PIC (Fourier-Bessel Particle-In-Cell).
+
+It tests the restart of the spin-tracker, with and without
+ionizer. Without ionization, the spin vectors are checked to be the
+same within some precision. For ionization, we check the simulations run.
```
RemiLehe:
```suggestion
same within some precision. For ionization, we check that the simulations run.
```

tests/test_spin_restart.py
```diff
@@ -0,0 +1,233 @@
...
+        os.path.join( temporary_dir, 'diags/hdf5') )
+    ts2 = OpenPMDTimeSeries(
+        os.path.join( temporary_dir, 'original_diags/hdf5') )
+
+    checked_species = [('electrons', 1e-3)]
+    if script_name == 'lwfa_script.py':
+        compare_spin_vectors( ts1, ts2, checked_species)
+    elif script_name == 'ionization_script.py':
+        checked_species += [('electrons from N', 0.2)]
+        compare_polarisations(ts1, ts2, checked_species, 'z')
```
RemiLehe, line 135: Could you add an `else` condition that raises an error, saying that the file is unknown?

tests/test_spin_restart.py
RemiLehe: Actually: it looks like it is doing a bit more in the case where ionization is on: it seems that this also checks that the mean polarization is similar enough. Could you update this comment accordingly?

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -52,6 +56,11 @@ class Ionizer(object):
     The implemented ionization model is the ADK model. The implementation
     is fully relativistic (i.e. it works in the boosted-frame as well).
 
+    This class also handles spin tracking. In case of an ion getting
+    ionized, the first electron will have the same spin as the ion, whereas
+    all subseqeuntly ionized electrons will have a random spin direction
```
RemiLehe:
```suggestion
    all subsequently ionized electrons will have a random spin direction
```

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -317,6 +329,38 @@ def handle_ionization( self, ion ):
             # (on GPU or GPU depending on `use_cuda`)
             generate_new_ids( elec, old_Ntot, new_Ntot )
 
+            # If spin tracking is enabled, generate also new spins
+            # (on GPU or GPU depending on `use_cuda`)
+            if ion.spin_tracker is not None:
+                if elec.spin_tracker is None:
+                    warnings.warn('\nSpin tracking is enabled for ion, '
+                                  'but not for target species!\n')
+                    continue
```
RemiLehe: In principle, this should never happen, given the check that you added in the constructor, no?
```suggestion
```

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -317,6 +329,38 @@ def handle_ionization( self, ion ):
...
+
+                if use_cuda:
+                    # Generate a set of random spins here
+                    #rand_sx = allocate_empty(new_Ntot-old_Ntot, True, np.float64)
+                    #rand_sy = allocate_empty(new_Ntot - old_Ntot, True, np.float64)
+                    #rand_sz = allocate_empty(new_Ntot - old_Ntot, True, np.float64)
```
RemiLehe:
```suggestion
```

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -317,6 +329,38 @@ def handle_ionization( self, ion ):
...
+                    copy_ionized_electron_spin_cuda[ batch_grid_1d,
+                                                     batch_block_1d ](
+                        N_batch, self.batch_size, old_Ntot,
+                        ion.Ntot, self.store_electrons_per_level,
+                        cumulative_n_ionized, i_level, ionized_from,
+                        elec.spin_tracker.sx, elec.spin_tracker.sy,
+                        elec.spin_tracker.sz, ion.spin_tracker.sx,
+                        ion.spin_tracker.sy, ion.spin_tracker.sz,
+                        rand_sx, rand_sy, rand_sz)
+                else:
```
RemiLehe, line 349: Is there a reason why `copy_ionized_electron_spin_cuda` and `copy_ionized_electron_spin_numba` look quite different? 
(i.e. the fact that `copy_ionized_electron_spin_numba` generates the random new spins inside the function, whereas `copy_ionized_electron_spin_cuda` does not)

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -28,13 +28,15 @@
 handled in batches of 10 particles, so that only the cumulative sum of the
 number of particles in each batch need to be performed.
 """
+import warnings
```
RemiLehe:
```suggestion
```
I suggested to remove the warnings further below.

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
...
+            The anomalous magnetic moment of the particle.
+            Default value is that of an electron.
+
+        spin_distr: str, optional
+            If 'fixed', all particles will have a fixed spin value
+            equal to s{x,y,z}_m.
+            If 'rand', the spin vectors will be random, but with an
+            ensemble average defined by one of the values of
+            s{x,y,z}_m. The first non-zero mean component will be
+            used, with order of preference being x,y,z, ie if sx_m!=0,
```
RemiLehe, line 573: Instead of picking the first non-zero mean, could we instead say that, when using `rand`, the user needs to make sure that only one of `sx_m`, `sy_m` or `sz_m` is non-zero. (and then we check that this is indeed the case in the code, and raise an error otherwise)

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
...
+            the generated spins will have an ensemble averages of
+            |sx|=sx_m, |sy|=0, |sz|=0, or if sz_m!=0, |sx|=0, |sy|=0
+            and |sz|=sz_m.
```
RemiLehe:
```suggestion
            <sx>=sx_m, <sy>=0, <sz>=0, or if sz_m!=0, <sx>=0, <sy>=0
            and <sz>=sz_m.
```

fbpic/particles/elementary_process/ionization/ionizer.py
```diff
@@ -152,7 +161,11 @@ def __init__(self, element, ionizable_species, target_species,
         for species in self.target_species:
             assert species.q == -e
             assert species.m == m_e
-
+            # Another sanity check: if spin tracking is enabled for
```
RemiLehe: Could we instead always ask the user to manually activate spin tracking, by raising an error here if it is not the case?

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
...
+                               anom=0.00115965218128, spin_distr='fixed'):
+        """
+        Activate spin tracking for this particle. This will
+        enable calculating the evolution of particle spin
+        vector throughout the simulation.
+
+        Note: spin tracking should be activated _before_ making
+        a species ionizable. Otherwise, spin tracking must be
+        manually activated for the target species of the
+        ionized electrons.
```
RemiLehe: See my comment in `ionizer.py`: I think that it might be cleaner to always explicitly ask the user to manually activate spin tracking, before activating ionization (and raise an error in `ionizer.py` otherwise)

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
...
+            used, with order of preference being x,y,z, ie if sx_m!=0,
+            the generated spins will have an ensemble averages of
+            |sx|=sx_m, |sy|=0, |sz|=0, or if sz_m!=0, |sx|=0, |sy|=0
+            and |sz|=sz_m.
+        """
+        # Warn about ionizer!
+        if self.ionizer is not None:
+            warnings.warn('\nIonizer already activated! Please make sure '
+                          'spin tracking is manually activated for all '
+                          'target species\n')
```
RemiLehe: See my comment in ionizer.py: I think that it might be cleaner to always explicitly ask the user to manually activate spin tracking, before activating ionization (and raise an error in ionizer.py otherwise).

We could also raise an error here, saying that spin tracking should be activated **before** the ionizer.
```suggestion
```

fbpic/particles/particles.py
RemiLehe: In the docstring, could you also mention the relationship to picking random points uniformly on a sphere? Does this only happen when `sx_m = sy_m = sz_m = 0`, and if yes, what is the random distribution in the case where one of these is non-zero?

fbpic/particles/particles.py
```diff
@@ -495,6 +511,61 @@ def make_ionizable(self, element, target_species,
...
+        vector throughout the simulation.
+
+        Note: spin tracking should be activated _before_ making
+        a species ionizable. Otherwise, spin tracking must be
+        manually activated for the target species of the
+        ionized electrons.
+
+        Parameters
+        ----------
+        sx_m: float
```
RemiLehe: Could you specify the units in which `sx_m`, `sy_m`, `sz_m` should be given?

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+    clipped to abs(1).
+    """
+    s1_th = min((0.3, 1 - abs(s1_m)))
+    s1 = s1_m * np.ones(Ntot) + s1_th * np.random.normal(size=Ntot)
+    s1[s1 > 1] = 1.  # keep within 1!
+    s1[s1 < -1] = -1.
+    s2 = 0.5 * np.random.normal(size=Ntot)
+    s3 = 0.5 * np.random.normal(size=Ntot)
+
+    ratio = np.sqrt((1 - s1 ** 2) / (s2 ** 2 + s3 ** 2))
```
RemiLehe: Could you explain a bit more what this code does, and how much different it is from randomly generating points that are uniformly distributed on a sphere (i.e. what `random_point_sphere_cpu` does)? In particular, in the limit where `s1_m` is very small, does this become very similar to what `random_point_sphere_cpu` does? (My impression is no, but maybe I am wrong here.)

Also: is there a paper that could be cited here, to explain the above method?

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        self.sx = self.sx.get()
+        self.sy = self.sy.get()
+        self.sz = self.sz.get()
+
+
+def make_random_spins(Ntot, s1_m):
+    """
+    Make a set of random spins. The component s1 will
+    have an average value of s1_m, with its distribution
+    width being up to 0.3. All spin components will be
```
RemiLehe: How was the value of `0.3` chosen? 
(See also the comment below on adding more info for what this code does.)

fbpic/particles/spin/spin_tracker.py
RemiLehe: Actually, here is a suggestion that generalizes the `random_point_sphere_cpu` and smoothly goes to uniform points on a sphere when `s1_m=0` ; would it make sense to use this function?
```python
def random_point_sphere_cpu(Ntot, s1_m):
    """
    Generates a random distribution of points on the surface of a sphere of radius 1, 
    such that the mean along z is `s1_m`. For s1_m=0, 
    points are uniformly distributed at the surface of the sphere
    """
    assert abs(s1_m) <= 1
    if abs(s1_m) == 1:
        z = s1_m * np.ones(Ntot) # z has a fixed value in this case
    elif s1_m == 0:
        z = np.random.uniform(-1, 1, Ntot) # random uniform
    else:
        # Generate z coord with a probability distribution e^(-alpha z) between -1 and 1
        
        # First find the value of alpha that correspond to s1_m
        # using the fact that the mean of the probability distribution e^(-alpha z)
        # is <z> = (1/alpha - 1/tanh(alpha))
        from scipy.optimize import fsolve
        sol = fsolve(lambda alpha: 1./alpha - 1./np.tanh(alpha) - s1_m, 0.5)
        alpha = sol[0]

        # Generate z with probability distribution e^(-alpha z)
        # by using the inverse CDF method     
        u = np.random.uniform(0, 1, Ntot)
        z = -1/alpha * np.log( np.exp(alpha)*(1-u) + np.exp(-alpha)*u )

    # azimuthal angle
    phi = np.random.uniform(0, 2 * pi, Ntot)
    # polar angle
    sin_theta = np.sin(np.arccos(z))
    # cartesian coords x and y
    x = sin_theta*np.cos(phi)
    y = sin_theta*np.sin(phi)
    return x, y, z
```
(Note that solving for `alpha` with `fsolve` could be done in the `__init__` function of the `spin_tracker` to avoid doing this everytime we call this function.)

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+    """
+    Class that stores and tracks particle spin vector.
+    """
+    def __init__(self, species, dt, sx_m=0., sy_m=0., sz_m=1.,
+                 anom=0., spin_distr='fixed'):
+        """
+        Initialize the SpinTracker class.
+
+        The length of each particle's spin vector is 1, and this
+        does not change during the push. During initialization,
```
RemiLehe, line 32: Saying that the spin of each particle is `1` is a bit ambiguous here, no?
(My understanding is that, when using the `fixed` method, the spin of the particles is not 1, but instead the floating-point value that the user provides.)

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        the spin vectors will be all set to the same value or
+        randomly generated to have the average values along all
+        axis specified by the user (see spin_distr below).
+
+        .. math::
+            \\frac{d\\boldsymbol{s}}{dt} = (\\boldsymbol{\\Omega}_T +
+             \\boldsymbol{\\Omega}_a) \\boldsymbol{s}
```
RemiLehe: This is supposed to be a vector product, correct?
```suggestion
             \\boldsymbol{\\Omega}_a) \\times \\boldsymbol{s}
```

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        sz_m: float (dimensionless), optional
+            The species-averaged average projection onto
+            the z-axis
+
+        anom: float, (dimensionless), optional
+            The anomalous magnetic moment of the particle,
+            given by :math:`a=(g-2)/2`, where :math:`g` is the
+            particle's g-factor.
+
+        spin_distr: str, optional
```
RemiLehe, line 88: I have the same comments here as for the similar docstring that is in `particles.py`.

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+            If 'fixed', all particles will have a fixed spin value
+            equal to s{x,y,z}_m.
+            If 'rand', the spin vectors will be random, but with an
+            ensemble average defined by one of the values of
+            s{x,y,z}_m. The first non-zero mean component will be
+            used, with order of preference being x,y,z, ie if sx_m!=0,
+            the generated spins will have an ensemble averages of
+            |sx|=sx_m, |sy|=0, |sz|=0, or if sz_m!=0, |sx|=0, |sy|=0
+            and |sz|=sz_m.
```
RemiLehe:
```suggestion
            <sx>=sx_m, <sy>=0, <sz>=0, or if sz_m!=0, <sx>=0, <sy>=0
            and <sz>=sz_m.
```

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+                 \\times \\frac{\\boldsymbol{E}}{c} \\right)
+
+        and
+
+        .. math::
+            \\boldsymbol{\\Omega}_a = a_e \\frac{q}{m}\\left(
+                 \\boldsymbol{B} -
+                 \\frac{\\gamma}{1+\\gamma}\\boldsymbol{\\beta}
+                 (\\boldsymbol{\\beta}\\cdot\\boldsymbol{B}) -
+                 \\boldsymbol{\\beta} \\times \\frac{\\boldsymbol{E}}{c} \\right)
```
RemiLehe, line 56: Could you also include the above equations in the docstring of `activate_spin_tracking`, in `particles.py`?

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        self.sx, self.sy, self.sz,                          n-1/2
+        ux_prev, uy_prev, uz_prev                           n-1/2
+        self.Ex, self.Ey, self.Ez,                          n
+        self.Bx, self.By, self.Bz,                          n
+        self.x, self.y, self.z                              n
+        self.ux, self.uy, self.uz                           n+1/2
+
+        NOTE!
+        Functions have not been modified to work with differing charge in
+        'ionizable' or 'plane' methods.
```
RemiLehe: I am not sure what is meant by the note here. Could you clarify?

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+                push_s_ioniz_gpu[dim_grid_1d, dim_block_1d](
+                    self.sx, self.sy, self.sz,
+                    self.ux_prev, self.uy_prev, self.uz_prev,
+                    self.ptcl.ux, self.ptcl.uy, self.ptcl.uz,
+                    self.ptcl.Ex, self.ptcl.Ey, self.ptcl.Ez,
+                    self.ptcl.Bx, self.ptcl.By, self.ptcl.Bz,
+                    self.ptcl.m, self.ptcl.Ntot, self.dt, self.anom,
+                    self.ptcl.ionizer.ionization_level)
+                # elif z_plane is not None:
+                # ... no implemented effect on spin precession for x_plane...
```
RemiLehe: I am not sure what this means ; should it be removed?
```suggestion
```

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+                push_s_ioniz_numba(self.sx, self.sy, self.sz,
+                                   self.ux_prev, self.uy_prev, self.uz_prev,
+                                   self.ptcl.ux, self.ptcl.uy, self.ptcl.uz,
+                                   self.ptcl.Ex, self.ptcl.Ey, self.ptcl.Ez,
+                                   self.ptcl.Bx, self.ptcl.By, self.ptcl.Bz,
+                                   self.ptcl.m, self.ptcl.Ntot,
+                                   self.dt, self.anom,
+                                   self.ptcl.ionizer.ionization_level)
+                # elif z_plane is not None:
+                # ... no implemented effect on spin precession for z_plane...
```
RemiLehe:
```suggestion
```

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+            elif self.sz_m != 0.:
+                sz, sx, sy = make_random_spins(Ntot, self.sz_m)
+            else:
+                # If the user does not pass anything, all spins
+                # are randomly oriented
+                sx, sy, sz = random_point_sphere_cpu(Ntot)
+
+        return sx, sy, sz
+
+    def generate_ionized_spins_gpu(self, Ntot):
```
RemiLehe: Could you add a brief docstring to this function?
In particular, it might good to say this generates the spins according to the specifications of `self.spin_distr`, `self.sx_m`, `self.sy_m`, `self.sz_m`.

fbpic/particles/spin/spin_tracker.py
```diff
@@ -0,0 +1,276 @@
...
+        self.sx_m = sx_m
+        self.sy_m = sy_m
+        self.sz_m = sz_m
+        self.sm = np.sqrt(sx_m ** 2 + sy_m ** 2 + sz_m ** 2)
+        self.anom = anom
+        self.dt = dt
+        self.spin_distr = spin_distr
+
+        # Store the species we perform spin tracking for
+        self.ptcl = species
```
RemiLehe: Could this be avoided, instead passing `species` explicitly as an argument to `store_previous_momenta` and `push_s`?
(This is mostly for stylistic preferences ; but if it is too much of a hassle, it's also fine to leave this as is.)

fbpic/particles/spin/inline_functions.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # Assume 1./gamma at midpoint can be calculated from (average)
+    # momentum at midpoint
+    inv_gamma = 1./math.sqrt( 1. + ux**2 + uy**2 + uz**2 )
+    # Define parameter: 1./(1+gamma)
+    inv_1pgamma = inv_gamma/(1. + inv_gamma)
+    """
+    NOTE! ALGORITHM WRITTEN IN TERMS OF REST FRAME FIELDS.
+    CHECK HOW THE FORM OF BMT EQUATION CHANGES FOR LORENTZ-BOOST.
+    """
```
RemiLehe: Hm, are you certain that the equations only work in a particular frame?
My guess (without checking explicitly) is that they should work in any frame (i.e. that the equations are invariant under a Lorentz transformation) and that they should therefore automatically work in boosted-frame simulations.
```suggestion
```

(Although the actual initialization of the spin may need to be modified when doing a boosted-frame simulation, so that we get the distribution in the boosted-frame, not the one that the user specifies in the lab frame.)

fbpic/particles/spin/inline_functions.py
```diff
@@ -0,0 +1,124 @@
...
+    # upsz = 2*tauz/(1. + tau2)
+    #
+    # New spin components
+    # sx_f = ( sx_i + sy_i*upsz - sz_i*upsy + (sz_i*taux - sx_i*tauz)*upsz
+    #          - (sx_i*tauy - sy_i*taux)*upsy )
+    # sy_f = ( sy_i + sz_i*upsx - sx_i*upsz + (sx_i*tauy - sy_i*taux)*upsx
+    #          - (sy_i*tauz - sz_i*tauy)*upsz )
+    # sz_f = ( sz_i + sx_i*upsy - sy_i*upsx + (sy_i*tauz - sz_i*tauy)*upsy
+    #          - (sz_i*taux - sx_i*tauz)*upsx )
+    # # -----------------------------------------------------------------------
```
RemiLehe:
```suggestion
```

fbpic/particles/spin/inline_functions.py
```diff
@@ -0,0 +1,124 @@
...
+
+    # ---------- Tamburini Method ---------------------------------------------
```
RemiLehe, line 50: Does this correspond to the method proposed here:
https://arxiv.org/pdf/2303.16966.pdf
If yes, could you add this reference as a comment?

RemiLehe (changes requested):
Thanks a lot for this impressive amount of work!!  I love that the code is high-quality, has several automated tests, and is well documented! 🚀 ✨ 

I added a number of suggestions and comments. Most of them are cosmetic. Some of the more significant comments are the following:
- One of the automated tests did not pass on GPU when I tried. Would you have time to fix it?
- The function that initialized the spin in the case `'random'` with a non-zero mean seemed a bit obscure to me. I suggested another function (see corresponding Python code in the inline comments).
- Did anyone test this in the boosted frame? Also: do you have a reference on how spins transform under a Lorentz transform? If yes, we could modify the initialization of the spin, in the case of a boosted-frame simulation, to make it consistent with the fact that the spins are initialized in a different Lorentz frame.

fbpic/particles/elementary_process/ionization/ionizer.py
pots007: Sounds good, I guess it's better, otherwise the user might get more than he bargained for.

tests/test_spin_restart.py
pots007: done!

tests/test_spin_ionization.py
pots007: What you suggested doesn't exactly work: the electrons that were ionized in the previous step would precess during the subsequent step, so the mean of the z spin component of _all_ electrons would <1. But I rewrote the loop, such that it now checks that <s_z> of _newly ionized_ electrons is correct.

tests/test_spin_planewave.py
pots007: Nope, that paper is for the eq. of motion. I added a note to point to where the spin evolution comes from.

fbpic/particles/spin/spin_tracker.py
pots007: Sorry, this was left from the first implementation of the code, when the charge state was hard-coded.

fbpic/particles/spin/spin_tracker.py
pots007: Thanks for pointing this out! I do not think it makes sense to have |s| < 1, so I've added a check to make sure 
the sum of squares of the components adds up to 1 if using 'fixed' method.

fbpic/particles/elementary_process/ionization/ionizer.py
pots007: I think I tried to have the cuda method generate spins inside the function as well, but it did not play nice with compile_cuda. Would that make sense?

fbpic/particles/spin/spin_tracker.py
pots007: Good idea! It was a lot easier than I thought as well :)

fbpic/particles/spin/spin_tracker.py
pots007: Turns out this was an oopsie: I implemented the `random_point_sphere_gpu` method directly in the ionizer. The code is much cleaner now, thanks for pointing this out.

fbpic/particles/spin/spin_tracker.py
pots007: Lovely idea, I've implemented that now!

fbpic/particles/spin/spin_tracker.py
pots007: This point is now moot with the improved generator you suggested.

fbpic/particles/particles.py
pots007: done

fbpic/particles/particles.py
pots007: 1st point: Yes, good idea, implemented that  now!
2nd point: Done!

tests/test_spin_planewave.py
pots007: No, I don't think it has enough precision. The anomalous moment makes a really small difference, becomes important at very high gamma though.

fbpic/particles/spin/inline_functions.py
pots007: I think you're right, the equations would work in another frame, it's 'just' a matter of transforming the spin components. I will try to look into it a bit more!

tests/test_spin_ionization.py
pots007: Note: the way I implemented checking only works for CPU sims, as the array order stays the same. When using GPU, the order of particles gets changed (ie new particles will end up at various indices, not at the end). So I've added that into the test as well.

tests/test_spin_ionization.py
pots007: I ran tests on GPU as well and could not generate this error! Bizarre. What file is the source of the error?

fbpic/particles/spin/inline_functions.py
pots007: done!

pots007:
Hey @RemiLehe , thanks a lot for the review and sorry it took me so long to get the changes done! I think I've addressed most of your points:
- the error I cannot reproduce. Could you let me know which fbpic file is the source of the error?
- the random spin generation is now based on your suggestion - that is much, much better
- Updated lots of docstrings
- Removed some weird circular function calls
- removed storing species as you suggested
- improved spin_ionization test
- etc

I have not yet done the boosted frame tests, I need a bit more time to think about that.

If you had time to check over the code again, it'd be great!

RemiLehe:
Great, thanks implementing these changes!
It looks like we both tried to fix the restart test at the same time ; in the end I pushed a slightly different fix.

Now all the tests are working on GPU on my side too, thanks for the fix!

I will take one last look soon and will then merge.

RemiLehe (approved):

PR #697, opened 2024-09-07
Added comment on ion initialisation in boosted frame

SchroederSa:
Added a note in the boosted frame example that ions are needed for correct modeling of typical setups


RemiLehe:
Thanks a lot for this PR!

PR #698, opened 2024-09-09
clarifying ion initialisation via initialize_ions flag in the simulat…

SchroederSa:
…ion class

I added a sentence on the need to specify also the ion density profile in the simulation class.


RemiLehe (approved):

RemiLehe:
Thanks!

PR #734, opened 2025-10-08
provide the non-zero `iteration` to lasy file

hightower8083:
This PR adds the optional argument to provide the `iteration` corresponding to the data in openpmd lasy file. In some cases non-zero or multiple iteration values maybe present and this argument is needed to handle them properly.


RemiLehe (approved):
Looks great! Thanks!

Issue #729, opened 2025-08-22
FBPIC not working after CUDA 13 driver upgrade


delaossa:
In August 2025 the Maxwell cluster at DESY upgrade to CUDA 13 drivers causing a malfunctioning of FBPIC.
After some big help from the IT experts at DESY (Frank Schluenzen), FBPIC is working again.

I will paste here the recipe that worked for me, in case is useful for the FPBIC user's community:

### Installing FBPIC

Installation script:

```bash
module purge
module load maxwell
module load cuda/13.0
module load mpi/openmpi-x86_64
module load mamba
. mamba-init

mamba create -n sim_env python=3.11 -y
mamba activate sim_env
pip install -U --no-cache-dir mpi4py --no-binary mpi4py
mamba install scipy h5py mkl -y
mamba install -c conda-forge -c nvidia numba-cuda cutensor nccl cuda-tools cuda-toolkit \
    cuda-nvvp cuda-nvtx cuda-nvrtc cuda-nvcc cuda-version=13.0 -y
pip install 'nvidia_nvjitlink>13.0'
pip install cupy-cuda13x
```

#### Changes to `FBPIC`

The `FBPIC` code also needed to be modified due to an incompatibility with `cuda.jit`:  
All instances of `inline=True` are changed to `False`, but this might not have been necessary for all of them.
You can make this change yourself or clone this specific branch:

```bash
git clone -b inline_false git@github.com:delaossa/fbpic.git
```

Then install normally `FBPIC` in your environment.

#### Running FBPIC

Example of a batch submission script to run `FBPIC` in parallel:

```bash
#!/bin/bash
#SBATCH --job-name=fbpic
#SBATCH --partition=mpa
#SBATCH --nodes=1
#SBATCH --ntasks-per-node=4
#SBATCH --constraint="A100-SXM4-40GB"
#SBATCH --time=1:00:00
#SBATCH --output=stdout
#SBATCH --error=stderr
#SBATCH --mail-type=END
#SBATCH --mail-user=<email>

unset LD_PRELOAD

# Activate environment for FBPIC
module purge
module load maxwell
module load cuda/13.0
module load mpi/openmpi-x86_64
module load mamba
. mamba-init
mamba activate sim_env

mpirun -n 4 --bind-to socket python fbpic_script.py
```

### Installing OPTIMAS

[...]

PR #735, opened 2025-10-09
Set `inline=False` for CUDA 13 compatibility

delaossa:
This PR implements a workaround by Frank Schlunzen (DESY IT support) to make `FBPIC` compatible with the new CUDA 13 drivers.
Check the issue here: https://github.com/fbpic/fbpic/issues/729

### DIFF
--- a/fbpic/lpa_utils/external_fields.py
+++ b/fbpic/lpa_utils/external_fields.py
@@ -134,7 +134,7 @@ class ExternalField( object ):
         self.cpu_func = cpu_compiler( func )
         if cuda_installed:
             # First create a device inline function
-            inline_func = cuda.jit( func, inline=True, device=True )
+            inline_func = cuda.jit( func, inline=False, device=True )
             # Then create a CUDA kernel and compile it the usual way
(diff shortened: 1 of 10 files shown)


RemiLehe:
Thanks a lot for contributing this change @delaossa 

Did you observe any regression in performance when removing `inline`? (e.g. do you have a benchmark simulation that you were able to run the previous CUDA version, and that you could re-run with this branch of FBPIC with CUDA 13?)

delaossa:
Hi @RemiLehe,
Sorry for the very long delay in my answer.
I didn't do proper performance tests, but just checking the running time of a production simulation on 1 GPU I didn't notice any significant performance drop: 19:33 (before) vs 19:37 (after).

AlexanderSinn:
According to
https://nvidia.github.io/numba-cuda/reference/kernel.html#numba.cuda.jit
https://numba.readthedocs.io/en/stable/developer/inlining.html
The inline option is more about how types are inferred when JIT compiling and not so much about whether an actual function call will be made in CUDA assembly.

The option for inlining at the CUDA assembly level is called `forceinline`.

PR #753, opened 2026-09-18
[WIP] Pydantic PICMI

ax3l:
Update the PICMI integration to the new pydantic implementation of PICMI.

## To Do

- [x] slop
- [ ] unslop (manual)
- [x] rebase after #752 is merged
- [ ] finalize PICMI version references once https://github.com/picmi-standard/picmi/pull/133 is merged an released

<details>
<summary>User-facing changes (Robot Generated)</summary>

**Requirements:** `picmistandard` 0.35.0 or later (the PICMI standard based on pydantic), which requires Python 3.11 or later and pydantic 2.13 or later. With an older `picmistandard`, importing `fbpic.picmi` raises an `ImportError` that asks to upgrade it.

**Changes of the PICMI standard itself** (for all PICMI codes):
- Arguments must be given by keyword, e.g. `picmi.FieldIonization(model='ADK', ionized_species=..., product_species=...)`.
- The input is validated: unknown arguments and invalid values raise a `pydantic.ValidationError` (a `ValueError`) instead of a `TypeError`/`AssertionError`. Arguments of other codes (e.g. `warpx_*`) are still ignored.
- Values are converted to the declared types, e.g. `charge_state=5` becomes `5.0` and NumPy arrays become lists.
- PICMI objects do not accept attributes that are not part of the standard (e.g. `species.tag = 1`), and the parameters of a `MultiSpecies` cannot be changed after it is created.
- `GriddedLayout(n_macroparticle_per_cell=...)` is deprecated in favor of `n_macroparticles_per_cell` (still accepted).

**Changes of the FBPIC PICMI bindings:**
- The FBPIC objects are created when the simulation is run (`sim.step()`) or when `sim.fbpic_sim` is accessed, instead of in each `add_*` method. The PICMI input can thus be given in any order, directly to `Simulation(...)`, or restored with `picmistandard.load(...)`. Input that is added after running is passed to FBPIC at the next `sim.step()`.
- Unsupported input (e.g. a laser profile other than `GaussianLaser`) is thus reported when the simulation is run, with a `ValueError` that describes it (instead of a bare `AssertionError`). The grid is still checked when the `Simulation` is created.
- Field ionization is now set up regardless of the order of `add_interaction` and `add_species`. Before, an interaction that was added after its species was silently ignored: the PICMI example `laser_acceleration_PICMI.py` does this, and thus ran without ionization. **Such scripts now include the ionization, which changes their results.**
- FBPIC uses the `particle_shape` of the species, or else the one of the `Simulation` (which PICMI defines as the default for the species). Since FBPIC uses the same shape for all species, these shapes must be the same. Before, the shape of the species was ignored: the species of the PICMI example `lpa_boostedframe_PICMI.py` request `'cubic'` shapes, and thus ran with `'linear'` shapes. **Such scripts now use the requested shapes, which changes their results.**
- Particle shapes can also be given as integers (`1` for `'linear'`, `3` for `'cubic'`), and the position of an injection plane as a scalar (its z position). Injection planes must be perpendicular to z, and are now also applied to Gaussian bunches (they were ignored).
- Particle diagnostics accept a single species, a `MultiSpecies`, or no species (all species).
- `species.fbpic_species` does not exist anymore: use `sim.get_fbpic_species(species)` instead (e.g. to call FBPIC's `track`). `sim.fbpic_sim` is read-only, and FBPIC does not modify the PICMI input anymore (the `charge`/`mass` of the species and the boundary conditions of the grid).
- The FBPIC-specific arguments are unchanged (e.g. `fbpic_show_progress=False`).

</details>


fbpic/picmi/simulation.py
```diff
@@ -33,43 +35,112 @@
...
+    # Create the FBPIC simulation object, once the PICMI input is validated
+    # (`model_post_init` replaces the `init` method of the pre-pydantic PICMI)
+    def model_post_init( self, context ):
+        super().model_post_init( context )
 
         # Get the grid
-        grid = self.solver.grid
-        if not type(grid) == PICMI_CylindricalGrid:
+        grid = None if self.solver is None else getattr(self.solver, 'grid', None)
+        if not isinstance(grid, PICMI_CylindricalGrid):
```
ax3l, line 128: Nice, a couple of these modernizations we can land separately already on `dev` to make the PR here smaller.

ax3l: #754

PR #754, opened 2026-09-18
PICMI: Use isinstance and super()

ax3l:
Replace the exact type comparisons in the PICMI bindings (e.g. `type(x) == PICMI_Species`) by `isinstance` checks, and call the methods of the parent class with `super()`.

This is a preparation for the pydantic-based PICMI (#753), where implementing codes derive their own classes from the standard ones, but it is fully compatible with the current `picmistandard<=0.34.0`.

Also fixes:
- the error message for an unsupported laser profile, which reported the type of the injection method (`PICMI_LaserAntenna`) instead of the laser
- the comment of `add_applied_field`


RemiLehe (approved):
Great! Thanks a lot for the clean PR and nice fix.

PR #755, opened 2026-09-21
PICMI: Fix AnalyticAppliedField

ax3l:
`PICMI_AnalyticAppliedField` always failed with `TypeError: The decorated object is not a function`: since 799d55f, a local `fieldfunc = None` shadowed the function that `exec` defined in the module globals.

The field function is now defined in its own namespace, with the user-defined variables and the functions/constants of `math` (e.g. `sin`, `pi`), and the expression is wrapped in parentheses.


RemiLehe (approved):

RemiLehe:
Thanks!

PR #756, opened 2026-09-21
PICMI: Fix Mirror

ax3l:
`PICMI_Mirror` failed with `TypeError` since the FBPIC `Mirror` takes `z_start`/`z_end` (27c4955), and with `AttributeError` in lab-frame simulations.

It is now passed as `[z_front_location, z_front_location + thickness]` (lab frame), with the thickness = max(`depth`, `number_of_cells` grid cells), or 2 cells by default, also in boosted frames.

Note: the FBPIC `Mirror` itself currently zeroes the cells just below `z_start` (and can drop a cell by truncation), so the mirror sits just below `z_front_location` until that is fixed separately.


fbpic/picmi/simulation.py
```diff
@@ -491,9 +491,30 @@ def add_applied_field(self, applied_field):
 
         if isinstance(applied_field, PICMI_Mirror):
             assert applied_field.z_front_location is not None
-            mirror = Mirror( z_lab=applied_field.z_front_location,
-                             n_cells=applied_field.number_of_cells,
-                             gamma_boost=self.fbpic_sim.boost.gamma0 )
+            # The mirror extends from `z_front_location` towards positive z.
+            # Its thickness is the maximum of `depth` and `number_of_cells`
+            # cells of the grid, or 2 cells if neither is set (default of the
+            # former FBPIC `Mirror`). `Mirror` takes lab-frame positions: in a
```
ax3l:
```suggestion
            # cells of the grid, or 2 cells if neither is set.
            # `Mirror` takes lab-frame positions: in a
```

ax3l:
Unrelated to this PICMI fix, Claude complained:

>   FBPIC's own Mirror class is also wrong, and #756 doesn't fix it:                                                                                                                                      
>  - Wrong side: it zeroes the cells just below z_start, not [z_start, z_end].                                                                                                                           
>  - Drops a cell: it truncates the cell count, so it zeroes one cell too few about half the time.                                                                                                       
>  - MPI: a mirror that crosses a subdomain boundary is only partly applied.                                                                                                                             

Separate PR for someone else :)

PR #758, opened 2026-09-21
Do not hide errors on CPU in catch_gpu_memory_error

ax3l:
On CPU, `OutOfMemoryError` is not imported, so any exception raised in a function decorated with `catch_gpu_memory_error` (e.g. `Ionizer.handle_ionization`) surfaced as `NameError: name 'OutOfMemoryError' is not defined`. Without CUDA, the decorator now returns the function unchanged.


fbpic/utils/printing.py
```diff
@@ -320,7 +320,13 @@ def catch_gpu_memory_error( f ):
 
     If a memory error occurs, this decorator prints a corresponding message
     and aborts the simulation (using MPI abort if needed)
+
+    When CUDA is not installed, `f` is returned unchanged.
     """
+    # Without CUDA, there is no GPU memory error to catch
+    # (and `OutOfMemoryError` is not imported)
+    if not cuda_installed:
+        return( f )
```
ax3l: Probably could be added in a bunch of functions in this file, but at least this one seems to be actively used in a bunch of CPU paths:
https://github.com/search?q=repo%3Afbpic%2Ffbpic+catch_gpu_memory_error&type=code

